PR #15, opened 2017-01-04
Safe load all the things

solarkennedy:


service_configuration_lib/__init__.py
```diff
@@ -59,9 +59,9 @@ def read_vip(vip_file):
 
 def load_yaml(fd):
     if yaml.__with_libyaml__:
-        return yaml.load(fd, Loader=yaml.CLoader)
+        return yaml.safe_load(fd, Loader=yaml.CLoader)
```
asottile: `safe_load` doesn't take a `Loader` argument

You want

```python
yaml.load(fd, Loader=yaml.CSafeLoader)
```

Here's some prior art: https://github.com/pre-commit/pre-commit-hooks/blob/5da199bb8d60f764c0f77a20b0a1dc3a7640bcdd/pre_commit_hooks/check_yaml.py#L8-L11

PR #26, opened 2019-06-27
Cache the absence of a file

solarkennedy:
I was tired of seeing thousands of calls to open "lb.yaml"


service_configuration_lib/__init__.py
```diff
@@ -95,6 +84,8 @@ def _read_yaml_file(file_name):
             if _use_yaml_cache:
                 _yaml_cache[file_name] = copy.deepcopy(data)
     except IOError:
```
vkhromov: `except FileNotFoundError:`, as `IOError` is an alias for `OSError` since some 3.x version and we probably don't want to ignore the file content after any even non-file-related error?

solarkennedy: Ah!

PR #29, opened 2019-07-12
Implement ConfigsFileWatcher to keep updated view on configuration files with pyinotify usage

billyevans:
Add YamlConfigsCachedView

### DIFF
--- a/service_configuration_lib/cached_view.py
+++ b/service_configuration_lib/cached_view.py
@@ -3,7 +3,7 @@ from collections import namedtuple
 import logging
 import os
 from pathlib import Path
-from typing import List
+from typing import Sequence
 from typing import Optional
 from typing import Tuple
 from abc import ABC
@@ -12,6 +12,7 @@ from abc import abstractmethod
 import pyinotify
 
 DEFAULT_SOA_DIR = '/nail/etc/services'
+DEFAULT_PROCESS_EVENTS_LIMIT = 100
 
 log = logging.getLogger(__name__)
 
@@ -65,10 +66,10 @@ class ConfigsFileWatcher:
         self,
         configs_view: BaseCachedView,
         configs_folder: str = DEFAULT_SOA_DIR,
-        services_names: List[str] = ['*'],
-        configs_names: Optional[List[str]] = None,
-        configs_suffixes: Optional[List[str]] = None,
-        exclude_folders_filters: List[str] = ['*.~tmp~'],
+        services_names: Sequence[str] = ['*'],
+        configs_names: Optional[Sequence[str]] = None,
+        configs_suffixes: Optional[Sequence[str]] = None,
+        exclude_folders_filters: Sequence[str] = ['*.~tmp~'],
     ) -> None:
         super().__init__()
         self.configs_view = configs_view
@@ -86,13 +87,18 @@ class ConfigsFileWatcher:
         log.info(f'max_user_instances: {pyinotify.max_user_instances.value}')
         log.info(f'max_user_watches: {pyinotify.max_user_watches.value}')
 
-    def process_events(self) -> None:
+    def process_events(self, limit: int = DEFAULT_PROCESS_EVENTS_LIMIT) -> None:
         """This method should be periodically invoked to process queued inotify
         events.  See the class docstring for more context.
+        Args:
+        :param limit: Optional, rough limit of events to process per call
         """
-        while self._notifier.check_events(timeout=0):
+        before_processed_count = self._processed_events_count
+        currently_processed = 0
+        while currently_processed >= 0 and currently_processed < limit and self._notifier.check_events(timeout=0):
             self._notifier.read_events()
             self._notifier.process_events()
+            currently_processed = self._processed_events_count - before_processed_count
 
     def close(self) -> None:
         if self._notifier is None:
@@ -104,6 +110,7 @@ class ConfigsFileWatcher:
         """Recreates state of ConfigsFileWatcher. Used as reaction on queue overflow in _EventHandler"""
         # close previous in case of re-creation
         self.close()
+        self._processed_events_count = 0
         handler = _EventHandler(cache=self)
         watch_manager = pyinotify.WatchManager()
         self._notifier = pyinotify.Notifier(
@@ -180,6 +187,9 @@ class ConfigsFileWatcher:
 
         return _ServiceConfig(service, config_name, config_suffix)
 
+    def _process_inotify_event(self):
+        self._processed_events_count += 1
+
     def __enter__(self):
         return self
 
@@ -193,12 +203,15 @@ class _EventHandler(pyinotify.ProcessEvent):
 
     # When acceptance testing, sometimes a move is seen as a create
     def process_IN_CREATE(self, event: pyinotify.Event) -> None:
+        self.cache._process_inotify_event()
         self.cache._maybe_add_path_to_cache(event.pathname)
 
     def process_IN_MOVED_TO(self, event: pyinotify.Event) -> None:
+        self.cache._process_inotify_event()
         self.cache._maybe_add_path_to_cache(event.pathname)
 
     def process_IN_DELETE(self, event: pyinotify.Event) -> None:
+        self.cache._process_inotify_event()
         self.cache._maybe_remove_path_from_cache(event.pathname)
 
     def process_IN_Q_OVERFLOW(self, event: pyinotify.Event) -> None:
--- a/service_configuration_lib/cached_view.py
+++ b/service_configuration_lib/cached_view.py
@@ -154,17 +154,9 @@ class ConfigsFileWatcher:
         relpath = os.path.relpath(path, self._configs_folder)
         filename = Path(relpath)
 
-        # when len == 2, thatss regular case e.g. service/config.yaml
-        # when len == 3, that's case with secrets folder, e.g. service/secrets/config.json
-        if len(filename.parts) == 2 or len(filename.parts) == 3:
-            service = filename.parts[0]
-            config_name = filename.stem
-            config_suffix = filename.suffix
-        else:
-            return None
-
-        if service == '.':
-            return None
+        service = str(filename.parts[0]) if len(filename.parts) > 1 else None
+        config_name = filename.stem
+        config_suffix = filename.suffix
 
         if not any((fnmatch.fnmatch(config_name, config) for config in self._configs_names)):
             return None
@@ -172,7 +164,7 @@ class ConfigsFileWatcher:
         if not any((fnmatch.fnmatch(config_suffix, suffix) for suffix in self._configs_suffixes)):
             return None
 
-        return str(service), config_name, config_suffix
+        return service, config_name, config_suffix
 
 
 class _EventHandler(pyinotify.ProcessEvent):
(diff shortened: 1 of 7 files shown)


service_configuration_lib/cached_view.py
```diff
@@ -0,0 +1,193 @@
...
+        :param path: the config file path e.g. `/nail/etc/services/foo/smartstack.yaml`
+        :return: a tuple of the service name, config and suffix e.g. (`foo`, `smartstack`, `yaml`),
+        or `None` if the path is invalid
+        """
+        relpath = os.path.relpath(path, self._configs_folder)
+        filename = Path(relpath)
+
+        # when len == 2, thatss regular case e.g. service/config.yaml
+        # when len == 3, that's case with secrets folder, e.g. service/secrets/config.json
+        if len(filename.parts) == 2 or len(filename.parts) == 3:
```
solarkennedy: This feels weird to me, and makes me think this shouldn't be in the cacher.

What if the raw filename was just returned and left to the caller to decide what to do with it?

Or, with all this config/service/file stuff, what if the caller was just responsible to provide a regex?

billyevans: Right, that's a bit weird.
The idea was to simplify service and config-names parsing from filename for library users.
It already passing full path, service_name and config_name to user's callbacks.
Perhaps I can have just assumption with len(filename.parts) > 1 and do same service_name and config_name extraction. Or do extraction when I can and put None for service_name into result when the files appeared on the top-level.

service_configuration_lib/cached_view.py
```diff
@@ -0,0 +1,185 @@
...
+        Returns `None` if path is invalid or we don't care about the file, e.g. not in self._config_names
+
+        :param path: the config file path e.g. `/nail/etc/services/foo/smartstack.yaml`
+        :return: a tuple of the service name, config and suffix e.g. (`foo`, `smartstack`, `yaml`),
+        or `None` if the path is invalid
+        """
+        relpath = os.path.relpath(path, self._configs_folder)
+        filename = Path(relpath)
+
+        service = str(filename.parts[0]) if len(filename.parts) > 1 else None
```
bplotnick: What's this line defending against? If there are less than 2 parts then I think the path is invalid?

service_configuration_lib/cached_view.py
```diff
@@ -0,0 +1,185 @@
...
+        config_name = filename.stem
+        config_suffix = filename.suffix
+
+        if not any((fnmatch.fnmatch(config_name, config) for config in self._configs_names)):
+            return None
+
+        if not any((fnmatch.fnmatch(config_suffix, suffix) for suffix in self._configs_suffixes)):
```
bplotnick: Not a big deal, but would it be simpler to make the suffixes (and maybe config names) exact matches rather than globs?

service_configuration_lib/cached_view.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def _maybe_remove_path_from_cache(self, path: str) -> None:
+        result = self._service_name_and_config_from_path(path)
+        if result is None:
+            return
+        service_name, config_name, config_suffix = result
+        log.info(f'{config_name}: Removing config for {service_name}')
+        self.configs_view.remove(path, service_name, config_name, config_suffix)
+
+    def _service_name_and_config_from_path(self, path: str) -> Optional[Tuple[str, str, str]]:
```
bplotnick: The API of returning the config and the suffix separately seems... weird to me. I guess it's not _that_ big of a deal. One option would be to bundle it into its own object/namedtuple would make it simpler. Or you could return a tuple of service and the `Path` object itself. Or you could ignore me 😅

service_configuration_lib/cached_view.py
```diff
@@ -0,0 +1,185 @@
...
+        events.  See the class docstring for more context.
+        """
+        while self._notifier.check_events(timeout=0):
+            self._notifier.read_events()
+            self._notifier.process_events()
+
+    def setup(self) -> None:
+        """Recreates state of ConfigsFileWatcher. Used as reaction on queue overflow in _EventHandler"""
+        handler = _EventHandler(cache=self)
+        watch_manager = pyinotify.WatchManager()
```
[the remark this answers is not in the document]
billyevans: Good catch, we actually need to close notifier, which will close inotify-fd.

service_configuration_lib/cached_view.py
billyevans: If len(parts) == 1 it will be top level file. Which probably is okay, we just can't extract service from it.

service_configuration_lib/cached_view.py
```diff
@@ -0,0 +1,185 @@
...
+    def log_inotify_constants() -> None:
+        log.info(f'max_queued_events: {pyinotify.max_queued_events.value}')
+        log.info(f'max_user_instances: {pyinotify.max_user_instances.value}')
+        log.info(f'max_user_watches: {pyinotify.max_user_watches.value}')
+
+    def process_events(self) -> None:
+        """This method should be periodically invoked to process queued inotify
+        events.  See the class docstring for more context.
+        """
+        while self._notifier.check_events(timeout=0):
```
vkhromov: It's probably better to limit amount of iterations for the loop, otherwise the current thread could stuck in this function for unspecified amount of time if there are a lot of changes if FS.

service_configuration_lib/yaml_cached_view.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def __init__(self):
+        super().__init__()
+        self.configs = defaultdict(dict)
+
+    def add(self, path: str, service_name: str, config_name: str, config_suffix: str) -> None:
+        try:
+            with open(path, encoding='utf-8') as fd:
+                self.configs[service_name][config_name] = yaml.load(fd, Loader=SafeLoader)
+        except OSError as exn:
```
vkhromov: Can we do something better than just printing warnings?
Because, let's be honest, nobody reads logs unless something horrible just happened.

May be immediately pass these exceptions to the caller of this code to deal with errors?
Or cache exceptions and re-raise then if somebody will try to access the corresponding cached entry?

service_configuration_lib/yaml_cached_view.py
```diff
@@ -0,0 +1,51 @@
...
+            log.warning(f'Error reading {path}: {exn}')
+        except yaml.YAMLError as exn:
+            log.warning(f'Error parsing {path}: {exn}')
+
+    def remove(self, path: str, service_name: str, config_name: str, config_suffix: str) -> None:
+        try:
+            del self.configs[service_name][config_name]
+        except KeyError:
+            log.warning(f'Config {config_name} not found for {service_name}')
```
vkhromov: the same as above about exceptions

service_configuration_lib/cached_view.py
```diff
@@ -0,0 +1,185 @@
...
+from typing import Tuple
+
+import pyinotify
+
+DEFAULT_SOA_DIR = '/nail/etc/services'
+
+log = logging.getLogger(__name__)
+
+
+class BaseCachedView:
```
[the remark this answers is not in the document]
billyevans: I resolved it once I fixed it in the code.

service_configuration_lib/yaml_cached_view.py
billyevans: This part is more like user-defined piece to react on filesystem changes. One of examples how to process maintain cache. Hence, it's up to user to write their own implementation instead of this one.
This particular implementation was written to behave in the same way as our current internal code.
I think that's better to leave it for now, once we will figure out another way of caching files and how to deal with exceptions we can add another it.

service_configuration_lib/yaml_cached_view.py
billyevans: See my comment above.

tests/cached_view_test.py
```diff
@@ -41,6 +47,15 @@ def test_exclude_filter_service_names_filtering(configs_file_watcher):
     assert not configs_file_watcher._exclude_filter('/foo/star_service')
 
 
+def test_stoping_notifier(configs_file_watcher):
```
bplotnick: nit: `s/stoping/stopping/`

service_configuration_lib/cached_view.py
```diff
@@ -77,16 +87,30 @@ def log_inotify_constants() -> None:
...
+    def process_events(self, limit: int = DEFAULT_PROCESS_EVENTS_LIMIT) -> None:
         """This method should be periodically invoked to process queued inotify
         events.  See the class docstring for more context.
+        Args:
+        :param limit: Optional, rough limit of events to process per call
         """
-        while self._notifier.check_events(timeout=0):
+        before_processed_count = self._processed_events_count
+        currently_processed = 0
+        while currently_processed >= 0 and currently_processed < limit and self._notifier.check_events(timeout=0):
```
bplotnick: is it possible for `currently_processed` to become negative? If so, when (and can we test that case 😄)?

tests/cached_view_test.py
```diff
@@ -81,6 +100,20 @@ def test_service_name_and_config_from_path_with_config_suffixes(configs_file_wat
...
 
 
+def test_configs_file_watcher_process_events_with_limit(configs_file_watcher):
+    configs_file_watcher._notifier.check_events.side_effect = [True, True, True, True, False]
+
+    configs_file_watcher._notifier.process_events.side_effect = configs_file_watcher._process_inotify_event
+
+    assert configs_file_watcher._processed_events_count == 0
+
+    configs_file_watcher.process_events(limit=2)
```
bplotnick: Is it worth testing whether `process_events` processes the next set of events when calling it again?

billyevans: Yes

service_configuration_lib/cached_view.py
```diff
@@ -0,0 +1,223 @@
...
+        configs_names: Optional[Sequence[str]] = None,
+        configs_suffixes: Optional[Sequence[str]] = None,
+        exclude_folders_filters: Sequence[str] = ['*.~tmp~'],
+    ) -> None:
+        super().__init__()
+        self.configs_view = configs_view
+        self._configs_folder = configs_folder
+        self._services_names = services_names
+        self._configs_names = None if configs_names is None else set(configs_names)
+        self._configs_suffixes = None if configs_suffixes is None else set(configs_suffixes)
```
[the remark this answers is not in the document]
billyevans: Yes, if it's None - library won't filter anything. If that's will be empty set/list, which actually makes not much sense it will filter everything, because nothing will match.

service_configuration_lib/cached_view.py
```diff
@@ -158,13 +179,22 @@ def _service_name_and_config_from_path(self, path: str) -> Optional[Tuple[str, s
...
+        return _ServiceConfig(service, config_name, config_suffix)
+
+    def _process_inotify_event(self):
+        self._processed_events_count += 1
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, err_type, err_val, err_tb):
+        self.close()
```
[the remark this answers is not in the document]
billyevans: I will add couple lines in the docs.

service_configuration_lib/cached_view.py
billyevans: Yes, I added tests

PR #36, opened 2019-12-04
fix reconfiguration bug when a folder is deleted

drmorr0:
Calling `setup` on a DELETE_SELF event causes the entire notifier to be recreated, but because we watch recursively, if a folder is deleted, a DELETE_SELF event occurs for the subfolder, which causes the notifier to be recreated.  Unfortunately, the old notifier was still (trying to) process events, and it would crash.

This change does two things: it delays the call to `setup()` until after the notifier has processed all events, and it makes sure that we only recreate the notifier in the event that the _root_ folder was deleted.  As far as I'm aware, there's no reason to recreate the notifier if a subfolder was deleted.

### DIFF
--- a/service_configuration_lib/cached_view.py
+++ b/service_configuration_lib/cached_view.py
@@ -76,6 +76,7 @@ class ConfigsFileWatcher:
     ) -> None:
         super().__init__()
         self.configs_view = configs_view
+        self._needs_reconfigure = False
         self._configs_folder = configs_folder
         self._services_names = services_names
         self._configs_names = None if configs_names is None else set(configs_names)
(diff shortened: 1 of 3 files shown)


drmorr0:
I manually confirmed that this fixes the repro case listed in CORESERV-8783

service_configuration_lib/cached_view.py
```diff
@@ -225,8 +230,9 @@ def process_IN_DELETE(self, event: pyinotify.Event) -> None:
...
     def process_IN_Q_OVERFLOW(self, event: pyinotify.Event) -> None:
         log.warning('Got queue overflow! Recreating watchers.')
-        self.cache.setup()
+        self.cache._needs_reconfigure = True
 
     def process_IN_DELETE_SELF(self, event: pyinotify.Event) -> None:
-        log.warning('Self folder was deleted! Recreating watchers.')
-        self.cache.setup()
+        if event.pathname == os.path.abspath(self.cache._configs_folder):
+            log.warning('Self folder was deleted! Recreating watchers.')
```
analogue: Can you embed the actual folder name in the message? Makes it a little more useful to uninitiated reader of the warning.

drmorr0: Done.

PR #39, opened 2019-12-19
Add initial spark_config function for generating the SPARK_OPTS environment variable

drmorr0:
This change does the following:

1. Add mypy support for this library (and fix a few related mypy issues)
2. Set up activate/deactivate functions for `aactivator`
3. Creates a `spark_config.py` file which can be imported and used to generate a dictionary containing the "right values" for the SPARK_OPTS environment variable.

Coverage fails with this issue, so I pin to <5.0: https://github.com/nedbat/coveragepy/issues/703

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,7 @@
+[mypy]
+python_version = 3.6
+check_untyped_defs = False
+warn_incomplete_stub = True
+follow_imports = silent
+ignore_missing_imports = True
+mypy_path = stubs
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -1,0 +1,116 @@
+import logging
+from typing import Any
+from typing import List
+from typing import Mapping
+from typing import MutableMapping
+from typing import Optional
+
+DEFAULT_SPARK_MESOS_SECRET_FILE = '/nail/etc/paasta_spark_secret'
+NON_CONFIGURABLE_SPARK_OPTS = {
+    'spark.master',
+    'spark.ui.port',
+    'spark.mesos.principal',
+    'spark.mesos.secret',
+    'spark.mesos.executor.docker.image',
+    'spark.mesos.executor.docker.parameters',
+    'spark.executorEnv.PAASTA_SERVICE',
+    'spark.executorEnv.PAASTA_INSTANCE',
+    'spark.executorEnv.PAASTA_CLUSTER',
+    'spark.executorEnv.SPARK_EXECUTOR_DIRS',
+    'spark.hadoop.fs.s3a.access.key',
+    'spark.hadoop.fs.s3a.secret.key',
+}
+log = logging.Logger(__name__)
+
+
+def get_mesos_spark_env(
+    spark_app_name: str,
+    spark_ui_port: str,
+    mesos_leader: str,
+    mesos_secret: str,
+    paasta_cluster: str,
+    paasta_pool: str,
+    paasta_service: str,
+    paasta_instance: str,
+    docker_img: str,
+    volumes: List[str],
+    user_spark_opts: Mapping[str, Any],
+    event_log_dir: Optional[str] = None,
+    needs_docker_cfg: bool = False,
+) -> Mapping[str, str]:
+    user_non_config_opts = set(user_spark_opts) & NON_CONFIGURABLE_SPARK_OPTS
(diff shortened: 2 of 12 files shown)


service_configuration_lib/spark_config.py
```diff
@@ -0,0 +1,106 @@
...
+    if spark_env['spark.eventLog.enabled'] == 'true':
+        if not spark_env.get('spark.eventLog.dir'):
+            if not event_log_dir:
+                raise ValueError('Asked for event logging but spark.eventLog.dir missing')
+            spark_env['spark.eventLog.dir'] = event_log_dir
+        log.info(f'Spark event logs available in {event_log_dir}')
+
+    spark_env['spark.mesos.executor.docker.parameters'] = 'cpus={}'.format(spark_env['spark.executor.cores'])
+    exec_mem = spark_env['spark.executor.memory']
+    if exec_mem[-1] != 'g' or not exec_mem[:-1].isdigit() or int(exec_mem[:-1]) > 32:
```
huadongliu: Thanks for making the change. Could you please relax the 32GB executor memory to warning? I have seen cases larger memory is needed.

service_configuration_lib/spark_config.py
```diff
@@ -0,0 +1,106 @@
...
+    'spark.mesos.secret',
+    'spark.mesos.executor.docker.image',
+    'spark.mesos.executor.docker.parameters',
+    'spark.executorEnv.PAASTA_SERVICE',
+    'spark.executorEnv.PAASTA_INSTANCE',
+    'spark.executorEnv.PAASTA_CLUSTER',
+    'spark.hadoop.fs.s3a.access.key',
+    'spark.hadoop.fs.s3a.secret.key',
+    'fs.s3a.access.key',
+}
```
stug: how did you come up with this list?  I am noticing that it doesn't fully match what we have in spark-run (which doesn't have `fs.s3a.access.key`) or spark_tools (which includes `fs.s3a.access.key` and `spark.app.name` (although I'm not sure why)).

drmorr0:
> lgtm. It would be nice to have a follow-up to update `paasta spark-run` to make use of the lib function introduced here.

@jangsutsr That is the plan, though it's going to be a while before we get there.  If we were to update `paasta spark-run` right now it would immediately switch over all of the production Tron jobs to this new code, which I don't want.  My rough plan of attack is a) push this code; b) iterate on services to make sure it works; c) migrate everything off `paasta spark-run` in Tron; d) replace the `paasta spark-run` code with this code.

service_configuration_lib/spark_config.py
drmorr0: Ummm I just kinda guessed.  I've taken `fs.s3a.access.key` out for now, and it "seems" OK to allow users to set the `spark.app.name`?

I also just added `spark.executorEnv.SPARK_EXECUTOR_DIRS` to this list, which it seems like maybe we don't want to let people set?

Maybe @huadongliu can comment on this list.

mypy.ini
```diff
@@ -0,0 +1,7 @@
+[mypy]
+python_version = 3.6
+check_untyped_defs = False
+warn_incomplete_stub = True
+follow_imports = silent
+ignore_missing_imports = True
+mypy_path = stubs
```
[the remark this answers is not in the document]
drmorr0: Done.

service_configuration_lib/spark_config.py
drmorr0: Yep, done.

drmorr0:
@huadongliu @stug I would like your opinion on whether this function should return a dict of Spark options or whether it should return the SPARK_OPTS-formatted string (e.g., `--conf X=Y --conf A=B...`)

stug:
Regarding the string of opts vs a dict, I vote you expose a method that returns the dict and a second one that can take the dict and turn it into the options string.  spark_tools only ever needs the dict, and spark-run needs the string.

One other thing to keep in mind is that mrjob uses `--jobconf` instead of `--conf`, so maybe we need a kwarg for the option string (if that doesn't feel too unclean).  It also pulls out the `spark.master` option because mrjob wants that specified separately -- separating creating the dict and the string would make that easier to handle.  (see https://github.com/Yelp/paasta/blob/v0.91.22/paasta_tools/cli/cmds/spark_run.py#L602 for where this happens).

PR #41, opened 2020-02-25
Create get_mesos_spark_auth_env helper function

jangsutsr:


setup.py
```diff
@@ -17,7 +17,7 @@
 
 setup(
     name='service-configuration-lib',
-    version='2.1.1',
+    version='3.0.0',
```
drmorr0: _technically_ this is backwards incompatible, but since we're the only consumers of this feature right now and we're still trying to get it to work right, I'd be sortof inclined to not do a major version bump here.  But it's up to you.

PR #43, opened 2020-03-02
Modifications to spark_opts getters

jangsutsr:
Changes include:
* Support mounting volumes to executor pods;
* Add force pull policy to executor pods;
* Verify file mode formats when passing volumes to executors on Mesos;
* Fix a bug where docker parameters is overwritten.

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -25,6 +25,7 @@ NON_CONFIGURABLE_SPARK_OPTS = {
     'spark.kubernetes.authenticate.caCertFile',
     'spark.kubernetes.authenticate.clientKeyFile',
     'spark.kubernetes.authenticate.clientCertFile',
+    'spark.kubernetes.container.image.pullPolicy',
 }
 K8S_AUTH_FOLDER = '/etc/spark_k8s_secrets'
 log = logging.Logger(__name__)
@@ -46,7 +47,13 @@ def get_mesos_spark_auth_env() -> Mapping[str, str]:
 
 def _check_non_configurable_spark_opts(user_spark_opts: Mapping[str, Any]) -> None:
     user_non_config_opts = set(user_spark_opts) & NON_CONFIGURABLE_SPARK_OPTS
-    if user_non_config_opts:
+    if (
+        user_non_config_opts or
+        any([
+            spark_opt.startswith('spark.kubernetes.executor.volumes.hostPath')
+            for spark_opt in user_spark_opts.keys()
+        ])
+    ):
         raise ValueError(f'The following Spark options are not user-configurable: {user_non_config_opts}')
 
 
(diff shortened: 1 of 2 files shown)


service_configuration_lib/spark_config.py
```diff
@@ -46,7 +47,13 @@ def get_mesos_spark_auth_env() -> Mapping[str, str]:
 
 def _check_non_configurable_spark_opts(user_spark_opts: Mapping[str, Any]) -> None:
     user_non_config_opts = set(user_spark_opts) & NON_CONFIGURABLE_SPARK_OPTS
-    if user_non_config_opts:
+    if (
+        user_non_config_opts or
+        any([
+            spark_opt.startswith('spark.kubernetes.executor.volumes.hostPath')
```
drmorr0: why not just add these to the set of non-configurable opts?

service_configuration_lib/spark_config.py
```diff
@@ -66,6 +73,12 @@ def get_mesos_spark_env(
 ) -> Mapping[str, str]:
     _check_non_configurable_spark_opts(user_spark_opts)
 
+    for volume in volumes:
+        volume_elements = volume.split(':')
+        if len(volume_elements) == 3 and volume_elements[-1] not in ('rw', 'ro'):
```
drmorr0: I vote that we call `.lower()` here; this way users don't have to worry about casing.  But it's still worth having the error in case of a typo.

service_configuration_lib/spark_config.py
```diff
@@ -93,19 +106,27 @@ def get_mesos_spark_env(
...
+    spark_env['spark.mesos.executor.docker.parameters'] = 'cpus={},{}'.format(
+        spark_env['spark.executor.cores'],
+        spark_env['spark.mesos.executor.docker.parameters'],
+    )
     _validate_spark_env(spark_env, event_log_dir)
 
     return spark_env
 
 
+def _generate_volume_name(volume):
```
drmorr0: I'm unconvinced that there's value in having this as a function, can't we just call `volume['containerPath']` directly where we need it?

service_configuration_lib/spark_config.py
```diff
@@ -93,19 +106,27 @@ def get_mesos_spark_env(
...
+
+
 def get_k8s_spark_env(
     spark_app_name: str,
     spark_ui_port: str,
     paasta_cluster: str,
     paasta_service: str,
     paasta_instance: str,
     docker_img: str,
+    volumes: List[Any],
```
drmorr0: nit: can we make the volumes a List of `TypedDict`s?

service_configuration_lib/spark_config.py
jangsutsr: Because these opts are dynamically generated therefore cannot be statically compared like the rest opts.

service_configuration_lib/spark_config.py
jangsutsr: In case we need to change the format of volume name to use. Current the option will look like `spark.kubernetes.executor.volumes.hostPath./nail/etc/beep.mount.path`, nothing wrong with it but later we might want to modify to increase readability, like `/nail/etc` to `nail-etc`.

service_configuration_lib/spark_config.py
jangsutsr: `TypedDict` cannot be used to declare argument type. I will use `Mapping` instead.

PR #44, opened 2020-03-10
Fix kubernetes volume name format

jangsutsr:
According to https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names, volume name must be a DNS_LABEL and unique within the pod.

In specific, volume name must:
    * contain at most 63 characters
    * contain only lowercase alphanumeric characters or ‘-’
    * start with an alphanumeric character
    * end with an alphanumeric character

Our current way of using path name is therefore invalid, as it contains illegal characters, like /.


drmorr0:
Please leave some comments/description in the PR so I can have some context on the changes being proposed :)

service_configuration_lib/spark_config.py
```diff
@@ -119,8 +119,19 @@ def get_mesos_spark_env(
...
+    Volume name must be a DNS_LABEL and unique within the pod. More info:
+    https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+
+    In specific, volume name must:
+    * contain at most 63 characters
+    * contain only lowercase alphanumeric characters or ‘-’
+    * start with an alphanumeric character
+    * end with an alphanumeric character
+    '''
+    return idx
```
drmorr0: Why do we have a function that literally just returns its argument?  Can't we just delete this and use the idx directly?

PR #45, opened 2020-04-20
Add labels for spark executors running in K8s

nisan1101:


service_configuration_lib/spark_config.py
```diff
@@ -26,6 +26,8 @@
     'spark.kubernetes.authenticate.clientKeyFile',
     'spark.kubernetes.authenticate.clientCertFile',
     'spark.kubernetes.container.image.pullPolicy',
+    'spark.kubernetes.executor.label.yelp.com/paasta_service',
+    'spark.kubernetes.executor.label.yelp.com/paasta_instance',
```
solarkennedy, line 30: Can you also add cluster?

https://github.com/Yelp/paasta/blob/5d3523b5ec57c9857137a7cb6cecebf9cd0931f7/paasta_tools/setup_kubernetes_cr.py#L265

solarkennedy:
I think I'm wrong about these labels?
https://github.com/Yelp/paasta/pull/2631/files

@keymone do we need both? or just paasta.yelp.com/*

keymone:
`paasta.yelp.com/name` is the correct scheme, `yelp.com/paasta_` is deprecated

nisan1101:
@solarkennedy @keymone  Im adding the labels to satisfy https://github.com/Yelp/synapse-tools/blob/master/src/synapse_tools/generate_container_ip_map.py#L71. Should I add both labels?

keymone:
no need to add both, just the new scheme

PR #48, opened 2020-06-03
Add parameter to read_extra_service_information that allows the user …

EvanKrall:
…to skip the call to copy.deepcopy, for performance.

### DIFF
--- a/service_configuration_lib/__init__.py
+++ b/service_configuration_lib/__init__.py
@@ -87,9 +87,13 @@ def read_dependencies(dependencies_file):
     return read_yaml_file(dependencies_file)
 
 
-def read_yaml_file(file_name):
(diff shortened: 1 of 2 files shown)


service_configuration_lib/__init__.py
```diff
@@ -87,9 +87,16 @@ def read_dependencies(dependencies_file):
     return read_yaml_file(dependencies_file)
 
 
-def read_yaml_file(file_name):
+def read_yaml_file(file_name, deepcopy=True):
+    """
+        Reads the given yaml file
```
tzhu-yelp, line 92: Absolutely thrilling docstring, not saying that docstrings are bad but there are probably more wirthy subjects

EvanKrall: lol. I think I was gonna add more here but got distracted.

PR #50, opened 2020-06-17
Set label and node-selector for kubernetes pods for spark jobs

88manpreet:
Kubernetes schedule uses node-selector to deduce the kubernetes pool from where the executor pods are allocated to run spark jobs. This is equivalent for spark on mesos was setting `spark.mesos.constraints` property.
Tested manually by running a spark job by setting the label and node-selector in the mesos box.
BATCHPROC-118 has more details.

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -134,6 +134,7 @@ def get_k8s_spark_env(
     docker_img: str,
     volumes: List[Mapping[str, str]],
     user_spark_opts: Mapping[str, Any],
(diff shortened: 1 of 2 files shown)


88manpreet:
My follow-up step will be to add `paasta_pool` param [here](https://github.com/Yelp/paasta/blob/master/paasta_tools/tron_tools.py#L250) after bumping and pinning the newer version.
Note that I set the param as mandatory instead of optional. I don't see any issue. Please feel free to let me know if I should make it an optional param

PR #54, opened 2020-10-02
combined spark functions from paasta and spark_tools

lydian:
This pull request combined both paasta spark-run/paasta.tron_tools/spark_tools.paasta into the same location, so that we only need to modify single location for any spark chagnes. 

The related changes based on this PR: 
- paasta: https://github.com/Yelp/paasta/pull/2930
- spark_tools: https://github.yelpcorp.com/python-packages/spark_tools/pull/13

# Testing
- I've modified the unit test to cover this
- I've also manually verified: 
  - Tested with the modified spark_tools: https://folium.yelpcorp.com/notebook/view/19383
  - Tested with paasta spark-run using https://sourcegraph.yelpcorp.com/packages/spark-on-paasta/-/blob/itest/s3.py, and it run succesfully: http://history.spark.paasta-pnw-devc.yelp/history/aba8bfb3-839f-41c1-9455-f40738e18a5c-8623/environment/
  - Tested with tron tools (mesos): https://fluffy.yelpcorp.com/i/1S4MWt9zj7nmbsWwWWVmsJR3cbW9QB5V.html
  - Tested with tron tools (kubernetes): https://fluffy.yelpcorp.com/i/7tGS7cnWCVK454MnpbBx50vH21Lk0RXl.html

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -4,6 +4,7 @@ import logging
 import os
 import time
 from typing import Any
+from typing import Dict
 from typing import List
 from typing import Mapping
 from typing import MutableMapping
@@ -12,7 +13,6 @@ from typing import Tuple
 from urllib.parse import urlparse
 
 import boto3
-import clusterman_metrics
 import ephemeral_port_reserve
 import requests
 import yaml
@@ -127,7 +127,7 @@ def _get_mesos_docker_volumes_conf(
     spark_opts: Mapping[str, str],
     extra_volumes: Optional[List[Mapping[str, str]]] = None,
     load_paasta_default_volumes: bool = False,
-) -> MutableMapping[str, str]:
+) -> Dict[str, str]:
     """return volume str to be configured for spark.mesos.executor.docker.volume
     if no extra_volumes and volumes_from_spark_opts, it will read from
     DEFAULT_PAASTA_VOLUME_PATH and parse it.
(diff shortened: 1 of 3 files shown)


service_configuration_lib/spark_config.py
```diff
@@ -162,66 +416,248 @@ def get_k8s_spark_env(
...
+        'cpus': num_cpus,
+        'mem': total_memory,
+        # rough guess since spark does not collect this information
+        'disk': total_memory,
+        'gpus': num_gpus,
+    }
+
+
+def generate_clusterman_metrics_entries(
+    clusterman_metrics,
```
kawaiwanyelp, line 592: Passing `clusterman_metrics` (i.e. a module) as an argument to where it is used is an unconventional way of having `service_configuration_lib` be able to use `clusterman_metrics` without needing to import it. When I suggested using an interface, I was envisioning more of a interface class that users (like `PaaSTA` and others) could initialize using `clusterman_metrics`. This interface pattern would be more in-line with being open-source.

That being said I think what you have here is okay, but I would include at minimum detailed comments explaining how this works and what `clusterman_metrics` is in this context, considering there is no import anywhere. I would also specify this is a Yelpism, as we make it pretty clear there are a number in this repo in the repo README.

kawaiwanyelp (changes requested):

service_configuration_lib/spark_config.py
lydian: Thanks, I added docstring and code example for how to init the clusterman_metrics before passing it.  Please let me know if there's anything you'd like to add.

lydian:
Due to the time constraints, i'll probably not going to add the changelog. I bumped the version and added documentation for the functions and parameters.

kawaiwanyelp (approved):

PR #57, opened 2020-12-10
Do not respond to the production target

tzhu-yelp:
This package is failing internal CI checks because it responds to the production target, doing nothing. Makefiles which respond to this target have other responsibilities.


nemacysts:
i somehow didn't see this and did this a couple months later in https://github.com/Yelp/service_configuration_lib/pull/59

PR #60, opened 2021-02-04
Automatically Configure TemporaryAWSCredentialsProvider if AWS Session Token Is Included in Session Credentials

smrojas:
https://github.com/Yelp/service_configuration_lib/pull/58 wasn't really the whole picture. When using a session token, a separate credentials provider must be included for hadoop s3a to work. In addition to [setting OS environ AWS values](https://hadoop.apache.org/docs/current/hadoop-aws/tools/hadoop-aws/index.html#Authenticating_via_the_AWS_Environment_Variables) in the [in the spark env](https://spark.apache.org/docs/latest/cloud-integration.html#authenticating) we must also declare the [TemporaryAWSCredentialsProvider](https://hadoop.apache.org/docs/current/hadoop-aws/tools/hadoop-aws/index.html#Using_Session_Credentials_with_TemporaryAWSCredentialsProvider) so they can be utilized. By default, the SimpleAWSCredentialsProvider is used, which doesn't work with session tokens.

Patched locally on https://github.com/Yelp/paasta/pull/3011/files. Verified that the spark conf flags get set automatically if the creds are passed in with a session token. After this change, the caller does not need to specify `--spark-args "spark.hadoop.fs.s3a.aws.credentials.provider=org.apache.hadoop.fs.s3a.TemporaryAWSCredentialsProvider"` for their spark job to work with temporary credentials. 

**Testing w/ Hadoop 3.0.1**
Verified that we can patch this locally with https://github.com/Yelp/service_configuration_lib/pull/58 and call pyspark with out needing to specify the provider. The spark-conf gets updated with the right provider, and pyspark interactive shell is open.
```
(py36-linux) srojas@dev2-uswest2adevc:~/paasta(u/srojas/SEC-13906-pass-session-token-to-spark-env⚡) » paasta spark-run -C pyspark --image docker-dev.yelpcorp.com/spark-
dev-srojas --aws-profile security --spark-args "spark.eventLog.dir=s3a://yelp-spark-event-logs-dev-us-west-2/srojas/"
/nail/home/srojas/paasta/paasta_tools/utils.py:1556: UserWarning: clog is unavailable
  warnings.warn("clog is unavailable")
{'spark.eventLog.dir': 's3a://yelp-spark-event-logs-dev-us-west-2/srojas/', 'spark.hadoop.fs.s3a.aws.credentials.provider': 'org.apache.hadoop.fs.s3a.TemporaryAWSCredent
ialsProvider'}
[...]

PR #61, opened 2021-02-08
Bump to Version 2.5.4

smrojas:
Follows up on https://github.com/Yelp/service_configuration_lib/pull/60, which I accidentally pushed the wrong version tag before pulling from latest on master.


smrojas:
No longer needed, just deleted the tag that I made before and re-pushed tags

PR #62, opened 2021-03-24
use spark.mesos.executor.memoryOverhead for memory calculation as well

talalryz:
I noticed that `spark.mesos.executor.memoryOverhead` was not being considered when calculating the memory requirements. This might be the possible cause of our issues

## testing done

make test green

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -643,11 +643,14 @@ def get_resources_requested(spark_opts: Mapping[str, str]) -> Mapping[str, int]:
     num_gpus = int(spark_opts.get('spark.mesos.gpus.max', 0))
 
     executor_memory = parse_memory_string(spark_opts.get('spark.executor.memory', ''))
+    requested_memory = spark_opts.get(
+        'spark.executor.memoryOverhead',
+    ) or spark_opts.get('spark.mesos.executor.memoryOverhead')
     # by default, spark adds an overhead of 10% of the executor memory, with a
     # minimum of 384mb
     memory_overhead: int = (
-        parse_memory_string(spark_opts['spark.executor.memoryOverhead'])
-        if spark_opts.get('spark.executor.memoryOverhead')
+        parse_memory_string(requested_memory)
+        if requested_memory
         else max(384, int(0.1 * executor_memory))
     )
     total_memory = (executor_memory + memory_overhead) * num_executors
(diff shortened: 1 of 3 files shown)


kawaiwanyelp (approved):
How are they different? They seem to be used in different places. A spark expert should confirm this makes sense.

jfongatyelp (approved):
Similar to @kawaiwanyelp I am not a spark expert so I still think a spark expert should approve, but I do see multiple job configs that configure this spark.mesos.executor.memoryOverhead as a significant % of executor_memory (e.g. 16g when executor memory is 32g), well above the default 10%, so this seems like a step in the right direction.

Some spark configurations also specify `spark.driver.memoryOverhead`, should this be included as well somewhere? Example: https://sourcegraph.yelpcorp.com/services/ad_attribution/-/blob/ad_attribution/batch/spark/calculate_campaign_driven_leads.py#L50:33

talalryz:
@jfongatyelp  for `spark.driver.memoryOverhead`, since we have the driver as the machine running the command (jupyterhub, or devbox), i don't think we should need to add that.

@kawaiwanyelp  `spark.mesos.executor.memoryOverhead` is the property recommended by [spark](https://spark.apache.org/docs/latest/running-on-mesos.html). I am actually not sure what uses `spark.executor.memoryOverhead` but too afraid to remove it

service_configuration_lib/spark_config.py
```diff
@@ -643,11 +643,14 @@ def get_resources_requested(spark_opts: Mapping[str, str]) -> Mapping[str, int]:
     num_gpus = int(spark_opts.get('spark.mesos.gpus.max', 0))
 
     executor_memory = parse_memory_string(spark_opts.get('spark.executor.memory', ''))
+    requested_memory = spark_opts.get(
+        'spark.executor.memoryOverhead',
+    ) or spark_opts.get('spark.mesos.executor.memoryOverhead')
```
88manpreet, line 648: We should also include `spark.kubernetes.memoryOverheadFactor` in the check too.
It seems to me by applying the standard definition of configuration overriding by child class, that `spark.mesos.executor.memoryOverhead` and `spark.kubernetes.memoryOverheadFactor` would override `spark.executor.memoryOverhead`.
To be on safe-side for resource allocation, should we take maximum of these 3?

88manpreet (approved):

PR #64, opened 2021-08-11
Update spark_config to handle k8s container required volumes for paasta spark-run

gcollura:
Change related to https://github.com/Yelp/paasta/pull/3174

## What changed
- [x] driver containers were missing essentials volumes when k8s was used as cluster manager, this review matches the behavior for the mesos CM.

PR #65, opened 2021-08-17
Change url for k8s cluster to be used for spark master

88manpreet:
Kubernetes cluster URL port seem to have been changed from 16443 to 6443.
As part of PAASTA-17390, the url format needs to be changed.

Testing:
make test

This should be a no-op in production as there is no production branch running spark on k8s job. 
Safe to rollout. 
Testing for same format for infrastage cluster has been done as part of COREML-2459


88manpreet:
Discarding in favor of https://github.com/Yelp/service_configuration_lib/pull/64#pullrequestreview-732073334

PR #66, opened 2021-08-18
Add GitHub actions

gcollura:
Taking inspiration from https://github.com/Yelp/task_processing/tree/master/.github/workflows and https://github.com/Yelp/paasta/pull/3126/files

How do I test it?


.github/workflows/ci.yml
```diff
@@ -0,0 +1,25 @@
...
+      matrix:
+        make_target: [test]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.6
+      # GHA won't setup tox for us and we use tox-pip-extensions for venv-update
+      - run: pip install tox==3.2 tox-pip-extensions==1.3.0
+      - run: make ${{ matrix.make_target }}
```
nemacysts, line 25:
```suggestion
    steps:
      - uses: actions/checkout@v2
      - uses: actions/setup-python@v2
        with:
          python-version: 3.6
      # GHA won't setup tox for us and we use tox-pip-extensions for venv-update
      - run: pip install tox==3.2 tox-pip-extensions==1.3.0
      - run: make test
```

is probably fine for now since we're only running one thing (seems like an unnecessary layer to have a matrix of one target)

.github/workflows/pypi.yml
```diff
@@ -0,0 +1,41 @@
...
+      matrix:
+        make_target: [test]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.6
+      # GHA won't setup tox for us and we use tox-pip-extensions for venv-update
+      - run: pip install tox==3.2 tox-pip-extensions==1.3.0
+      - run: make ${{ matrix.make_target }}
```
nemacysts, line 23:
```suggestion
    steps:
      - uses: actions/checkout@v2
      - uses: actions/setup-python@v2
        with:
          python-version: 3.6
      # GHA won't setup tox for us and we use tox-pip-extensions for venv-update
      - run: pip install tox==3.2 tox-pip-extensions==1.3.0
      - run: make test
```
same here :)

setup.py
```diff
@@ -17,7 +17,7 @@
 
 setup(
     name='service-configuration-lib',
-    version='2.5.6',
+    version='2.5.7',
```
nemacysts, line 20:
```suggestion
    version='2.5.6',
```
we normally bump the version outside of a PR (and then tag that commit in which we bumped the verison)

nemacysts (commented):
you'll need the pypi token to be added before merging, but looks good!

nemacysts (approved):

PR #68, opened 2021-08-31
[COREML-2570] fix k8s volumes being mounted twice

gcollura:
## What changed
Avoid getting spark-run stuck on attempting to mount the same volume twice. Add tests.

## Testing
Build paasta venv with updated library from local path.
```
./paasta_tools/cli/cli.py spark-run --cluster-manager kubernetes -s spark --cmd 'spark-submit /work/integration_tests/s3.py'
```
Output: https://fluffy.yelpcorp.com/i/MqQ4kTV56z1dV80wCGMpKMC7BCdjX0gG.html

### DIFF
--- a/tests/spark_config_test.py
+++ b/tests/spark_config_test.py
@@ -1,3 +1,4 @@
+import itertools
 import json
 import os
 from unittest import mock
@@ -165,7 +166,9 @@ class TestGetSparkConf:
 
     @pytest.fixture
     def mock_existed_files(self, mock_paasta_volumes):
-        existed_files = [v.split(':')[0] for v in mock_paasta_volumes] + ['/host/file1', '/host/file2', '/host/file3']
+        existed_files = [v.split(':')[0] for v in mock_paasta_volumes] + [
+            '/host/file1', '/host/file2', '/host/file3', '/etc/pki/spark', '/etc/group', '/etc/passwd',
+        ]
         with mock.patch('os.path.exists', side_effect=lambda f: f in existed_files):
             yield existed_files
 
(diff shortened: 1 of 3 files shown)


tests/spark_config_test.py
```diff
@@ -220,6 +223,45 @@ def test_get_mesos_docker_volumes_conf(
...
+                {'hostPath': '/host/file2', 'containerPath': '/containter/file2', 'mode': 'RO'},
+                {'hostPath': '/host/paasta1', 'containerPath': '/container/paasta1', 'mode': 'RO'},
+            ],
+        ],
+    )
+    @pytest.mark.usefixtures('mock_existed_files')
+    def test_get_k8s_docker_volumes_conf(self, volumes):
+        expected_volumes = {}
+
+        def _get_k8s_volume(host_path, container_path, mode, count=itertools.count()):
```
88manpreet, line 240: I don't think we should duplicate an api definition inside the test file. 
Can this be imported and used? Probably, we can define this outside `_get_k8s_docker_volumes_conf` api.

88manpreet:
Curious to know how did you notice this? Were there duplicate mounts in the logs?
Was the impact performance degradation?

88manpreet (approved):

gcollura:
> Curious to know how did you notice this? Were there duplicate mounts in the logs?
> Was the impact performance degradation?

It only happened in services/spark, because the /etc/pki/spark path was mounted twice. This prevented the whole Spark job to bootstrap successfully due to errors thrown by k8s master. Mesos does the same type of checking.

88manpreet (approved):

tests/spark_config_test.py
```diff
@@ -220,6 +224,45 @@ def test_get_mesos_docker_volumes_conf(
...
+                {'hostPath': '/host/file2', 'containerPath': '/containter/file2', 'mode': 'RO'},
+                {'hostPath': '/host/paasta1', 'containerPath': '/container/paasta1', 'mode': 'RO'},
+            ],
+        ],
+    )
+    @pytest.mark.usefixtures('mock_existed_files')
+    def test_get_k8s_docker_volumes_conf(self, volumes):
+        expected_volumes = {}
+
+        _get_k8s_volume = functools.partial(spark_config._get_k8s_volume_hostpath_dict, count=itertools.count())
```
88manpreet, line 250: Nice usage of partial!

gcollura: it was either this or using contexts, since it's a private method I went with the former.

PR #69, opened 2021-09-01
Creating TypeDict for ServiceInfo

cuza:
This PR creates a TypeDict for `service_configuration_lib.read_service_configuration_from_dir()` to make more clear how magically `smartstack`/`port`/`monitoring`/`deploy`/`data`/`dependencies`, etc. are loaded in `InstanceConfigDict`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 importlib-resources==1.5.0
 pyinotify==0.9.6
 pyyaml >= 3.0
+mypy-extensions==0.4.1
--- a/service_configuration_lib/__init__.py
+++ b/service_configuration_lib/__init__.py
@@ -22,6 +22,8 @@ import os
 import socket
 import sys
 from typing import Mapping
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -1,3 +1,4 @@
 importlib-resources==1.5.0
 pyinotify==0.9.6
 pyyaml >= 3.0
+mypy-extensions==0.4.1
```
nemacysts, line 4: this repo is in a weird state requirements-wise ;_;

that said, this also needs to be referenced in https://github.com/Yelp/service_configuration_lib/blob/master/setup.py#L27 (but with no pin, or at most a lower bound on whatever mypy-extensions added TypedDict)

requirements.txt will only be used for building a debian package, but the setup.py bit i linked is what pip will use for actually installing this as a normal python package

mattmb:
Cleaning up and closing some very old PRs. Please re-open or nudge me if you’re still planning to work on this.

PR #70, opened 2021-09-15
[COREML-2591] Fix issue with long label names in k8s

gcollura:
[Follow the RFC indicated in the wiki](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names) for label names.

Manual testing: https://fluffy.yelpcorp.com/i/CpTn5S74K85sDVGmxPzX3BsM5gBkx2QL.html#L1-2,L83-107,L1547-1556,L1643-1644
with master, the batch fails because the values set through the tron environment variables was exceeding 63 chars, with this branch, the batch successfully bootstrap and runs.

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -1,10 +1,10 @@
+import base64
 import functools
 import hashlib
 import itertools
 import json
 import logging
 import os
-import re
 import time
 from typing import Any
 from typing import Dict
@@ -450,9 +450,10 @@ def _get_k8s_spark_env(
 ) -> Dict[str, str]:
     # RFC 1123: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names
     # technically only paasta instance can be longer than 63 chars. But we apply the normalization regardless.
-    _paasta_cluster = _get_k8s_resource_name_prefix(paasta_cluster)[:DEFAULT_K8S_LABEL_LENGTH]
-    _paasta_service = _get_k8s_resource_name_prefix(paasta_service)[:DEFAULT_K8S_LABEL_LENGTH]
-    _paasta_instance = _get_k8s_resource_name_prefix(paasta_instance)[:DEFAULT_K8S_LABEL_LENGTH]
+    # NOTE: this affects only k8s labels, not the pod names.
+    _paasta_cluster = _get_k8s_resource_name_limit_size_with_hash(paasta_cluster)
+    _paasta_service = _get_k8s_resource_name_limit_size_with_hash(paasta_service)
+    _paasta_instance = _get_k8s_resource_name_limit_size_with_hash(paasta_instance)
     spark_env = {
         'spark.master': f'k8s://https://k8s.{paasta_cluster}.paasta:6443',
         'spark.executorEnv.PAASTA_SERVICE': paasta_service,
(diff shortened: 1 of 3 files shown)


chelspea (approved):

service_configuration_lib/spark_config.py
```diff
@@ -460,19 +467,23 @@ def _get_k8s_spark_env(
...
+        'spark.kubernetes.executor.label.paasta.yelp.com/cluster': _paasta_cluster,
         'spark.kubernetes.node.selector.yelp.com/pool': paasta_pool,
         'spark.kubernetes.executor.label.yelp.com/pool': paasta_pool,
         **_get_k8s_docker_volumes_conf(volumes),
     }
     return spark_env
 
 
+def _get_k8s_resource_name_prefix(resource_name: str):
+    return re.sub('-+', '-', re.sub('[^a-z0-9\\-]', '-', resource_name.strip().lower()))
```
88manpreet, line 484: Can you add some examples for this?
It looks like this expression is:
* Replacing all consecutive --'s to a single - and mostly any alpha-numeric characters except \ to -?
Is adding \\ required?

I also noticed that the failing label includes "paasta-spark-etl-runner" as prefix in the beginning. Do you think, it would be useful to truncate or replace it with shorter string to start with. That way it is less likely to have conflicts with 2 services having same label name after truncation? 
We know spark-etl is always on paasta and is a runner. So just maybe, 'spark-etl' or even 'etl'.
Thoughts?
Adding a docstring here would be useful. Took me a while to understand what is happening here :) 

I think adding a warning log line would be helpful here for before and after resource-name string if it changed or truncated.


Probably not in this review, but a unit-test would be useful

88manpreet:
For this pr, I would definitely want to get it reviewed by someone for CI, (maybe luisp). So that they are in loop, that this is happening. 
I am also think they might have seen this problem in the past too or other k8s services.

service_configuration_lib/spark_config.py
gcollura: >  mostly any alpha-numeric characters except \ to -
including `\`, I'm actually just following the rules for those resource names, I haven't encountered these cases yet. I might actually add `_` which is getting replaced to `-` but it may cause confusion.

> I also noticed that the failing label includes "paasta-spark-etl-runner" as prefix in the beginning.
No it's this case, the failing label is not in the log unfortunately (that's what made this bug tricky), but it's in the values that are set for `spark.kubernetes.executor.label.yelp.com/paasta_instance` and `spark.kubernetes.executor.label.paasta.yelp.com/instance` which in Tron follow the rule of `tron_job_name.tron_job_action`, set [here](https://sourcegraph.yelpcorp.com/Yelp/paasta@91a8540d9c2e5c5fc50b1048639b50f5d231e912/-/blob/paasta_tools/cli/cmds/spark_run.py#L395-398)

> That way it is less likely to have conflicts with 2 services having same label name after truncation?
good point, this might still apply in this case. Although these labels are not set to uniquely identify a job (more like filtering down a job) with the initial substring part it should still be able to filter adequately. In the long run, I'm planning after this change to introduce new tron specific labels, like `spark.kubernetes.executor.label.tron.yelp.com/job_name` and `spark.kubernetes.executor.label.tron.yelp.com/job_action`, but that's part of another change.

> I think adding a warning log line would be helpful here for before and after resource-name string if it changed or truncated.
these labels are purely used in case you want to filter down jobs with `kubectl-pnw-devc -n paasta-spark get pods -o wide -l paasta.yelp.com/instance=instance_name`. 

>  Adding a docstring here would be useful. Took me a while to understand what is happening here :)
good point! will add!

service_configuration_lib/spark_config.py
```diff
@@ -446,6 +448,11 @@ def _get_k8s_spark_env(
     volumes: Optional[List[Mapping[str, str]]],
     paasta_pool: str,
 ) -> Dict[str, str]:
+    # RFC 1123: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names
+    # technically only paasta instance can be longer than 63 chars. But we apply the normalization regardless.
+    _paasta_cluster = _get_k8s_resource_name_prefix(paasta_cluster)[:DEFAULT_K8S_LABEL_LENGTH]
+    _paasta_service = _get_k8s_resource_name_prefix(paasta_service)[:DEFAULT_K8S_LABEL_LENGTH]
+    _paasta_instance = _get_k8s_resource_name_prefix(paasta_instance)[:DEFAULT_K8S_LABEL_LENGTH]
```
nemacysts, line 455: might be nice to be consistent with how we shorten things in paasta (I recently fixed this for Tron since I had forgotten about this limit in https://github.com/Yelp/paasta/pull/3208/files) - https://github.com/Yelp/paasta/blob/f65ecfa6d6b5d3074951c640e4bf2b8bc16b826d/paasta_tools/kubernetes_tools.py#L409:5

and while we're looking at labels in this PR: it's probably worth going through y/paasta-workload-contract and making sure that the other labels specified there are being set (as well as some of the other k8s settings)

nemacysts (approved):

88manpreet (approved):

PR #71, opened 2021-11-09
[COREML-2694] automatically compute and set k8s executor limit cores and batch size

gcollura:
## Context

We want to automatically set `spark.kubernetes.allocation.batch.size` and  `spark.kubernetes.executor.limit.cores`. The former will help scaling up clusters for accommodating the executors, the latter will prevent certain Spark jobs to starve each others cpu pool.

## Testing
`make test` updated and passing.

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -36,6 +36,7 @@ DEFAULT_EXECUTOR_CORES = 2
 DEFAULT_EXECUTOR_INSTANCES = 2
 DEFAULT_EXECUTOR_MEMORY = '4g'
 DEFAULT_K8S_LABEL_LENGTH = 63
+DEFAULT_K8S_BATCH_SIZE = 512
 
 
 NON_CONFIGURABLE_SPARK_OPTS = {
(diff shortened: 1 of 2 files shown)


chelspea (approved):

service_configuration_lib/spark_config.py
```diff
@@ -36,6 +36,7 @@
 DEFAULT_EXECUTOR_INSTANCES = 2
 DEFAULT_EXECUTOR_MEMORY = '4g'
 DEFAULT_K8S_LABEL_LENGTH = 63
+DEFAULT_K8S_BATCH_SIZE = 512
```
88manpreet, line 39: Maybe change this to 1024 or higher (Basically more consistent with clusterman config of max_weight_to_add)?
Also, add the link here for the max_weight_to_add config. They kind of tie together.
With 512 and cores as 1. We only utilize 512 of 2500 available capacity to scale.

88manpreet:
`kubectl describe pod | grep Limit` would be a nice way to test before and after the deplopyed job.

service_configuration_lib/spark_config.py
gcollura: > add the link here for the max_weight_to_add config
it would be a private corp link in a public repo :/ 

> Maybe change this to 1024 or higher (Basically more consistent with clusterman config of max_weight_to_add)?
we can go back later and bump the value if needed, or if necessary certain batches can set up a higher value.

This type of config will help 90-95th percentile of batches in our system, if not more.

88manpreet: Makes sense. gfi

PR #72, opened 2021-12-11
[COREML-2800] set default spark.scheduler.maxRegisteredResourcesWaitingTime for kubernetes batches

gcollura:
Ticket: [COREML-2800](https://jira.yelpcorp.com/browse/COREML-2800)
As per ticket, we're going to set the default max time as 10min (instead of 15min, it's been working well for some batches I've been testing) plus 2 seconds for each executor that need to be instantiated. I also updated testing accordingly.

## Testing

`make test`: updated and passes ✅ 
Manual test:
```
 ▲ py37-linux ~pg/paasta pip install ../service_configuration_lib
 △ py37-linux ~pg/paasta ./paasta_tools/cli/cli.py spark-run --cluster pnw-devc --service ad_bidding_transform --pool stable_batch --cluster-manager kubernetes --cmd 'spark-submit /code/virtualenv_run/bin/spark_etl_runner.py --feature-config /code/batch/campaign_info_daily/job_configs/ad_shutoff_and_traffic_integration.yaml --env-config-path /nail/srv/configs/ad_bidding_transform.yaml --team-name ad_market_dynamics --no-notification --start-date 2021-09-16 --end-date 2021-09-16 --publish-path s3a://yelp-emr-dev-us-west-2/gcoll/k8s_darklaunch/ ' --spark-args ' spark.executor.instances=8 spark.executor.memory=16g spark.executor.cores=4 spark.driver.memory=8g spark.driver.cores=5'
```
image

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -38,6 +38,7 @@ DEFAULT_EXECUTOR_MEMORY = '4g'
 DEFAULT_K8S_LABEL_LENGTH = 63
 DEFAULT_K8S_BATCH_SIZE = 512
 DEFAULT_RESOURCES_WAITING_TIME_PER_EXECUTOR = 2  # seconds
+DEFAULT_CLUSTERMAN_OBSERVED_SCALING_TIME = 15  # minutes
 
 
 NON_CONFIGURABLE_SPARK_OPTS = {
@@ -293,9 +294,13 @@ def _adjust_spark_requested_resources(
             str(min(executor_instances, DEFAULT_K8S_BATCH_SIZE)),
         )
         user_spark_opts.setdefault('spark.kubernetes.executor.limit.cores', str(executor_cores))
+        waiting_time = (
+            DEFAULT_CLUSTERMAN_OBSERVED_SCALING_TIME +
+            executor_instances * DEFAULT_RESOURCES_WAITING_TIME_PER_EXECUTOR // 60
+        )
         user_spark_opts.setdefault(
             'spark.scheduler.maxRegisteredResourcesWaitingTime',
-            str(10 + executor_instances * DEFAULT_RESOURCES_WAITING_TIME_PER_EXECUTOR // 60) + 'min',
+            str(waiting_time) + 'min',
         )
 
     if max_cores < executor_cores:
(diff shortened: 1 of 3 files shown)


service_configuration_lib/spark_config.py
```diff
@@ -292,6 +293,10 @@ def _adjust_spark_requested_resources(
             str(min(executor_instances, DEFAULT_K8S_BATCH_SIZE)),
         )
         user_spark_opts.setdefault('spark.kubernetes.executor.limit.cores', str(executor_cores))
+        user_spark_opts.setdefault(
+            'spark.scheduler.maxRegisteredResourcesWaitingTime',
+            str(10 + executor_instances * DEFAULT_RESOURCES_WAITING_TIME_PER_EXECUTOR // 60) + 'min',
```
88manpreet, line 298: In my experiments I have seen 15 min. But I think 10 is also ok, as it's a soft limit.

chelspea, line 298: I think for safety maybe just do `ceil` rather than `\\`

88manpreet: Question: If the user is providing this config, does it overrides it?
Maybe we should take higher of the two, if the user specifies too?

88manpreet:
Only one concern added above. lgtm otherwise. Looking forward to manual test-run.

service_configuration_lib/spark_config.py
gcollura: What do you mean by safety? `ceil` and `//` (or `floor`) doesn't make much difference in this scenario.

88manpreet: > It might be misleading, no? It's not generally the approach that we follow for other settings in this library.
I am not sure about this. 
Example case: For some reason, the calculated waiting time is not enough.
User wants to specify config to higher value, as it's important for them to get the desired executor count.
They won't be able specify it, right? As it will be overridden everytime.
This makes the waiting-time config option non-configurable. 

It seems to me a trivial change to pick-up higher of given and calculated value. 

What are the other user configs which are overridden always?

88manpreet: Add 10 as a constant value macro at the top.
Something like, CLUSTERMAN_OBSERVED_SCALING_TIME.

> it gets to 15min pretty quickly with the right amount of executors anyway.
Not sure, I follow this.

88manpreet (approved):
lgtm
A manual test with user config overriding the calculated config would be nice, even after pushing the code change.

PR #74, opened 2022-02-15
[COREML-2697] handle missing kubernetes config values for mass migration

gcollura:
Infer when missing:
* `spark.executor.instances`: basically `spark.cores.max / spark.executor.cores`
* `spark.mesos.executor.memoryOverhead` -> `spark.executor.memoryOverhead`

Update test to reflect this use case.

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -286,7 +286,7 @@ def _adjust_spark_requested_resources(
         executor_instances = max_cores / executor_cores
     elif cluster_manager == 'kubernetes':
         # TODO(gcoll|COREML-2697): Consider cleaning this part of the code up
-        # once we mesos is not longer around at Yelp.
+        # once mesos is not longer around at Yelp.
         if 'spark.executor.instances' not in user_spark_opts:
             executor_instances = int(user_spark_opts.get('spark.cores.max', str(DEFAULT_MAX_CORES))) // executor_cores
             user_spark_opts['spark.executor.instances'] = str(executor_instances)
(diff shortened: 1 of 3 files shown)


chelspea (approved):
LGTM

gcollura:
> Is there any usage of spark.mesos.*... type args aside from the pullImage one that needs translating?

Not as far as I've seen during the migration.

service_configuration_lib/spark_config.py
```diff
@@ -285,6 +285,17 @@ def _adjust_spark_requested_resources(
         max_cores = int(user_spark_opts.setdefault('spark.cores.max', str(DEFAULT_MAX_CORES)))
         executor_instances = max_cores / executor_cores
     elif cluster_manager == 'kubernetes':
+        # TODO(gcoll|COREML-2697): Consider cleaning this part of the code up
+        # once mesos is not longer around at Yelp.
+        if 'spark.executor.instances' not in user_spark_opts:
+            executor_instances = int(user_spark_opts.get('spark.cores.max', str(DEFAULT_MAX_CORES))) // executor_cores
+            user_spark_opts['spark.executor.instances'] = str(executor_instances)
```
88manpreet, line 292: Having a Warning message here, would be helpful. This is to build the practice having this way as  preferred way and for consistency.
Indicating that it is recommended to add `spark.executor.instances=xyz` in configs.

service_configuration_lib/spark_config.py
```diff
@@ -285,6 +285,17 @@ def _adjust_spark_requested_resources(
...
+        if (
+            'spark.mesos.executor.memoryOverhead' in user_spark_opts and
+            'spark.executor.memoryOverhead' not in user_spark_opts
+        ):
+            user_spark_opts['spark.executor.memoryOverhead'] = user_spark_opts['spark.mesos.executor.memoryOverhead']
```
88manpreet, line 297: Same here.

service_configuration_lib/spark_config.py
nemacysts: can we just fix this for users once we're 100% off of mesos?

PR #78, opened 2022-05-23
Set AWS creds via Spark Config

jsleight:
In HADOOP-18233 I learned that Spark v3.2 has a race condition which requires that aws creds be set via the spark config instead of the aws environment variables.

We probably want to delay merging this PR into PaaSTA until we modify our Spark services to not rely on the AWS env variables.

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -894,7 +894,7 @@ def _emit_resource_requirements(
             writer.send((metric_key, int(time.time()), required_quantity))
 
 
-def _get_spark_hourly_cost(
+def get_spark_hourly_cost(
     clusterman_metrics,
     resources: Mapping[str, int],
     cluster: str,
@@ -937,7 +937,7 @@ def send_and_calculate_resources_cost(
     cluster = spark_conf['spark.executorEnv.PAASTA_CLUSTER']
     app_name = spark_conf['spark.app.name']
     resources = get_resources_requested(spark_conf)
-    hourly_cost = _get_spark_hourly_cost(clusterman_metrics, resources, cluster, pool)
+    hourly_cost = get_spark_hourly_cost(clusterman_metrics, resources, cluster, pool)
     _emit_resource_requirements(
         clusterman_metrics, resources, app_name, spark_web_url, cluster, pool,
     )


talalryz (approved):
lgtm!

88manpreet:
2 points:
1. Security concern: We have display spark configs enabled. During the test run in the driver logs, were the creds visible? I would believe so. In that case, it is a security concern. I think that's why env creds were preferred.

2.  > We probably want to delay merging this PR into PaaSTA until we modify our Spark services to not rely on the AWS env variables.
I could not follow the concern around this. If the creds are both through AWS env and through spark configs, would the race condition still exist? I would assume, it should just be redundant. 

Implementation lgtm. Thanks for adding explanation for adding both for s3a and s3 configs.

jsleight:
> 1. Security concern: We have display spark configs enabled. During the test run in the driver logs, were the creds visible? I would believe so. In that case, it is a security concern. I think that's why env creds were preferred.

No, it shows them as redacted. Same as what shows up in the spark UI's environment tab.

> 2. > We probably want to delay merging this PR into PaaSTA until we modify our Spark services to not rely on the AWS env variables.
>    > I could not follow the concern around this. If the creds are both through AWS env and through spark configs, would the race condition still exist? I would assume, it should just be redundant.

I would have thought they were redundant too, but for reasons which I do not understand they aren't.  If the env variables exist then we get the warnings.  So this PR makes it so the env variables do not get set at all.

Since some of our other spark utilities expected the env variables they also need to be changed, and we should change them to accept credentials from either the env variables or the spark configs.  Then once that change is done we can merge this PR (otherwise all those spark services would break)

jsleight:
@88manpreet e.g., like this
```
spark.hadoop.fs.s3.access.key=*********(redacted)
spark.hadoop.fs.s3.impl=org.apache.hadoop.fs.s3a.S3AFileSystem
spark.hadoop.fs.s3.secret.key=*********(redacted)
spark.hadoop.fs.s3.session.token=*********(redacted)
spark.hadoop.fs.s3a.access.key=*********(redacted)
spark.hadoop.fs.s3a.aws.credentials.provider=org.apache.hadoop.fs.s3a.TemporaryAWSCredentialsProvider
spark.hadoop.fs.s3a.multiobjectdelete.enable=true
spark.hadoop.fs.s3a.secret.key=*********(redacted)
spark.hadoop.fs.s3a.session.token=*********(redacted)
```

PR #82, opened 2022-07-28
MLCOMPUTE-36 | change default cachedExecutorIdleTimeout and expose get_dra_configs as public

CaptainSame:
The value of cachedExecutorIdleTimeout increased to 15 min = 900 sec for batches which use their cache at a later point in the application.
Exposed the get_dra_configs method as public to update dra configs in paasta: https://github.com/Yelp/paasta/pull/3408

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -41,7 +41,7 @@ DEFAULT_RESOURCES_WAITING_TIME_PER_EXECUTOR = 2  # seconds
 DEFAULT_CLUSTERMAN_OBSERVED_SCALING_TIME = 15  # minutes
 DEFAULT_SQL_SHUFFLE_PARTITIONS = 128
 DEFAULT_DRA_EXECUTOR_ALLOCATION_RATIO = 0.8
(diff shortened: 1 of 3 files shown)


88manpreet (approved):
lgtm


Please add the description.

service_configuration_lib/spark_config.py
```diff
@@ -271,6 +271,12 @@ def _get_dra_configs(spark_opts: Dict[str, str]) -> Dict[str, str]:
         spark_opts, 'spark.dynamicAllocation.cachedExecutorIdleTimeout',
         str(DEFAULT_DRA_CACHED_EXECUTOR_IDLE_TIMEOUT),
     )
+    log.warning(
+        f'\nSetting spark.dynamicAllocation.cachedExecutorIdleTimeout as {DEFAULT_DRA_CACHED_EXECUTOR_IDLE_TIMEOUT}. '
```
88manpreet, line 275: Nit: Maybe also add a reason of increasing. Something like, "If your job is performing bad as the cached data was lost, try increasing the cachedTimeoutValue.

Separate: It would also be nice to know the error one gets or an indication in logs, on how one can that missed cached blocks were expected to be used later. An experiment with low value of timeout and observing could a starting point.

PR #83, opened 2022-08-02
Add cluster-manager=local support.

jsleight:
Add support for local cluster-manager.  Will follow this up with a PR to paasta once this is merged.

I happened to notice that the `test_leaders_get_spark_conf_kubernetes` was mis-spelled and as a consequence pytest wasn't actually picking it up as a test to run.  So had to do a bit of fixing for that test as well.

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -762,7 +762,7 @@ def get_spark_conf(
 ) -> Dict[str, str]:
     """Build spark config dict to run with spark on paasta
 
-    :param cluster_manager: which manager to use, value supported: [`mesos`, `kubernetes`]
+    :param cluster_manager: which manager to use, must be in SUPPORTED_CLUSTER_MANAGERS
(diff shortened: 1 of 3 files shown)


88manpreet:
Overall looks good to me.
Nit: `    :param cluster_manager: which manager to use, value supported: [`mesos`, `kubernetes`]`. Maybe also add `local`? 

I believe some of the most common manual test cases would be added once this is released and tested with paasta.
 
Note: With `mesos`, spark.cores.max is used. With `kubernetes` spark.executor.instances are used. Probably needs to be check, what is the right config used by spark in local mode.

88manpreet (approved):

jsleight:
> Note: With `mesos`, spark.cores.max is used. With `kubernetes` spark.executor.instances are used. Probably needs to be check, what is the right config used by spark in local mode.

local mode doesn't use either of these configs, `local[XXX]` will use XXX numbers of threads.  Of course other configs are fine, they just get ignored

PR #84, opened 2022-08-11
MLCOMPUTE-439 | modified executor instance calculation in absence of spark.cores.max

CaptainSame:
- Currently, if a user provides `spark.executor.cores` but does not provide `spark.cores.max` and `spark.executor.instances`, `spark.executor.instances` will be calculated based on the (default value of `spark.cores.max` //  `spark.executor.cores`). This can lead to executors being set as 0 or a low value. If the user has not specified `spark.cores.max` and `spark.executor.instances`, this should signify that the user wants at least the default number of executors. The changes in the code are corresponding to this logic.

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -433,9 +433,13 @@ def _adjust_spark_requested_resources(
             # spark.cores.max provided, calculate based on (max cores // per-executor cores)
             if 'spark.cores.max' in user_spark_opts:
                 executor_instances = int(user_spark_opts['spark.cores.max']) // executor_cores
-            # spark.executor.instances and spark.cores.max not provided, the executor instances should at least
-            # be equal to DEFAULT_EXECUTOR_INSTANCES.
+                log.warning(
+                    f'spark.cores.max should be replaced and the exact value of spark.executor.instances '
+                    f'should be provided in --spark-args',
+                )
             else:
+                # spark.executor.instances and spark.cores.max not provided, the executor instances should at least
+                # be equal to DEFAULT_EXECUTOR_INSTANCES.
                 executor_instances = max(DEFAULT_MAX_CORES // executor_cores, DEFAULT_EXECUTOR_INSTANCES)
             user_spark_opts['spark.executor.instances'] = str(executor_instances)
 
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -416,6 +416,23 @@ def _append_aws_credentials_conf(
     return spark_opts
 
 
+def compute_executor_instances_k8s(user_spark_opts: Dict[str, str]) -> int:
+    executor_cores = int(user_spark_opts.get('spark.executor.cores', DEFAULT_EXECUTOR_CORES))
+    if 'spark.cores.max' in user_spark_opts:
+        # spark.cores.max provided, calculate based on (max cores // per-executor cores)
+        executor_instances = (int(user_spark_opts['spark.cores.max']) // executor_cores)
+        log.warning(
+            f'spark.cores.max should no longer be provided and should be replaced '
+            f'by the exact value of spark.executor.instances in --spark-args',
+        )
+    else:
+        # spark.executor.instances and spark.cores.max not provided, the executor instances should at least
+        # be equal to DEFAULT_EXECUTOR_INSTANCES.
+        executor_instances = max(DEFAULT_MAX_CORES // executor_cores, DEFAULT_EXECUTOR_INSTANCES)
+
+    return executor_instances
+
+
 def _adjust_spark_requested_resources(
     user_spark_opts: Dict[str, str],
     cluster_manager: str,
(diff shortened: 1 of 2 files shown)


service_configuration_lib/spark_config.py
```diff
@@ -430,14 +430,16 @@ def _adjust_spark_requested_resources(
...
         # once mesos is not longer around at Yelp.
         if 'spark.executor.instances' not in user_spark_opts:
-
-            executor_instances = int(user_spark_opts.get(
-                'spark.cores.max',
-                str(DEFAULT_MAX_CORES),
-            )) // executor_cores
+            # spark.cores.max provided, calculate based on (max cores // per-executor cores)
+            if 'spark.cores.max' in user_spark_opts:
+                executor_instances = int(user_spark_opts['spark.cores.max']) // executor_cores
```
88manpreet, line 435: We need to go in the direction, that `spark.cores.max` should not be provided.

Add a warning, that this should be replaced with config, `spark.executor.instances`. 
I think once we have a warning log, maybe in future we could override this config for all jobs.

service_configuration_lib/spark_config.py
```diff
@@ -430,14 +430,16 @@ def _adjust_spark_requested_resources(
...
+            # spark.executor.instances and spark.cores.max not provided, the executor instances should at least
```
88manpreet, line 436: Nit: Either move this comment on top of `if` or inside `else`

service_configuration_lib/spark_config.py
```diff
@@ -430,14 +430,16 @@ def _adjust_spark_requested_resources(
...
+            # be equal to DEFAULT_EXECUTOR_INSTANCES.
+            else:
+                executor_instances = max(DEFAULT_MAX_CORES // executor_cores, DEFAULT_EXECUTOR_INSTANCES)
             user_spark_opts['spark.executor.instances'] = str(executor_instances)
 
-            if user_spark_opts['spark.executor.instances'] == str(DEFAULT_EXECUTOR_INSTANCES):
+            if user_spark_opts['spark.executor.instances'] <= str(DEFAULT_EXECUTOR_INSTANCES):
                 log.warning(
                     f'spark.executor.instances not provided. Setting spark.executor.instances as '
```
88manpreet, line 444: This log message is confusing to me. 
If `spark.executor.instances` is greater than DEFAULT_EXECUTOR_INSTANCES it could be still be possible that it's not provided, right? Example: spark.cores.max = 100 and spark.executor.cores=2, no?

service_configuration_lib/spark_config.py
88manpreet: I think the `spark.executor.instances` can still come up to be 0, right?
Example: spark.cores.max=1, spark.executor.cores=2
Maybe in the end set: spark.executor.instaces = max(calculated_one, 1)?

CaptainSame: Added a warning. Regarding the second comment, if the executor instances come up to be zero when both spark.cores.max and spark.executor.cores are provided, it will throw an error on line 478

service_configuration_lib/spark_config.py
CaptainSame: Yes, it is still possible but this was added to warn in case we are using the DEFAULT_EXECUTOR_INSTANCES. Should I add it in every case? Also, this warning does not make sense if DRA is enabled but at this point, it is not clear if DRA will be enabled due to mass rollout executor range. How should I proceed?

88manpreet: hmm This is getting confusing with lots of if/else conditions. I think eventually we would want all batches to provide either `executor.instances` or `DRA configs for instances`. Anything else should be removed or warned. 

Remove that warning. 

Maybe just revisit the logic in order: 
if `executor.instances` provided use that:
elif `cores.max` is provided, use that with warning:
else use default value; 

Keep it as a separate api to calculate instance config.

service_configuration_lib/spark_config.py
```diff
@@ -416,6 +416,25 @@ def _append_aws_credentials_conf(
...
 
+def compute_executor_instances_k8s(user_spark_opts: Dict[str, str]) -> int:
+    if 'spark.executor.instances' in user_spark_opts:
+        return int(user_spark_opts['spark.executor.instances'])
+
+    executor_cores = int(user_spark_opts.get('spark.executor.cores', DEFAULT_EXECUTOR_CORES))
+    if 'spark.cores.max' in user_spark_opts:
+        # spark.cores.max provided, calculate based on (max cores // per-executor cores)
+        executor_instances = (int(user_spark_opts['spark.cores.max']) // executor_cores)
+        log.warning(
```
88manpreet, line 427:
```suggestion
```

service_configuration_lib/spark_config.py
```diff
@@ -416,6 +416,25 @@ def _append_aws_credentials_conf(
...
+            f'spark.cores.max should no longer be provided and should be replaced '
```
88manpreet, line 428:
```suggestion
```

service_configuration_lib/spark_config.py
```diff
@@ -416,6 +416,25 @@ def _append_aws_credentials_conf(
...
+            f'by the exact value of spark.executor.instances in --spark-args',
```
88manpreet, line 429:
```suggestion
```

service_configuration_lib/spark_config.py
```diff
@@ -416,6 +416,25 @@ def _append_aws_credentials_conf(
...
+        )
```
88manpreet, line 430:
```suggestion
```

service_configuration_lib/spark_config.py
```diff
@@ -416,6 +416,25 @@ def _append_aws_credentials_conf(
...
+    else:
+        # spark.executor.instances and spark.cores.max not provided, the executor instances should at least
+        # be equal to DEFAULT_EXECUTOR_INSTANCES.
+        executor_instances = max(DEFAULT_MAX_CORES // executor_cores, DEFAULT_EXECUTOR_INSTANCES)
+    return executor_instances
```
88manpreet, line 435:
```suggestion
      log.warning(
            f'spark.cores.max should no longer be provided and should be replaced '
            f'by the exact value of spark.executor.instances in --spark-args: "spark.executor.instances={executor_instances}"',
        )
    return executor_instances
```

88manpreet (approved):
lgtm. Just a minor comment on moving the log warning somewhere else.

PR #85, opened 2022-08-24
make Spark on Mesos use old spark shuffle fetch protocol

nanflasted:
Spark 3 introduced a new shuffle protocol (see [Apache Spark's migration guide](https://spark.apache.org/docs/3.2.0/core-migration-guide.html#upgrading-from-core-24-to-30) for more descriptions). Upon upgrading to Spark 3.2 it seems like we are running into issue with this new protocol on Mesos clusters, therefore we force spark to use the old fetch protocol to make spark on mesos work.

This doesn't seem to be a problem for spark on k8s so this is not added to the k8s configs. 

`make test` passed.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
-importlib-resources==1.5.0
+importlib-resources==5.4.0
 pyinotify==0.9.6
 pyyaml >= 3.0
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -1,3 +1,3 @@
-importlib-resources==1.5.0
+importlib-resources==5.4.0
```
nanflasted, line 1: this requirement works (as in `make test` passes) now on both internal and public PyPI.

jsleight (approved):

PR #86, opened 2022-08-30
Update signalfx link to use y/spark-metrics

88manpreet:
Use y/spark-metrics which is more catered towards spark metrics than the status-quo y/paasta-status


CaptainSame (approved):
LGTM

PR #87, opened 2022-09-16
adding PAASTA_SOA_DIR env var to set custom yelpsoa-configs directory

cuza:
Adding PAASTA_SOA_DIR env var to set custom yelpsoa-configs directory.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,7 @@ from setuptools import setup
 
 setup(
     name='service-configuration-lib',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -17,7 +17,7 @@
 
 setup(
     name='service-configuration-lib',
-    version='2.11.2',
```
nemacysts, line 20: probably want to bump this outside of this PR

PR #91, opened 2022-10-14
MLCOMPUTE-481 | handled spark config creation for service account credentials

CaptainSame:
- Spark is going to be run using a service account's credentials in Jupyter
- Added configs to handle service account spark creation

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -633,6 +633,7 @@ def _get_k8s_spark_env(
     docker_img: str,
     volumes: Optional[List[Mapping[str, str]]],
     paasta_pool: str,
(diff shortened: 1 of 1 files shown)


service_configuration_lib/spark_config.py
```diff
@@ -666,6 +664,20 @@ def _get_k8s_spark_env(
         'spark.kubernetes.executor.label.yelp.com/owner': 'core_ml',
         **_get_k8s_docker_volumes_conf(volumes),
     }
+    if service_account_name:
```
88manpreet, line 667:
```suggestion
    if service_account_name is not None:
```

88manpreet (approved):

PR #92, opened 2022-11-02
Remove showConsoleProgress flag

88manpreet:
This flag would show the live progress of a spark job on the console (driver logs).
This is really useful for debugging.
However, as part of tron jobs it is noisy, and not useful to parse for debugging purposes.
This can be added on-demand as part of adhoc jobs.


jsleight (commented):
lgtm but need to fix test failures

PR #94, opened 2022-11-21
get_spark_conf now sets additional spark.sql configs related to default partitioning.

jsleight:
- Set some additional configs to be inline with the shuffle partition numbers
- Increased the multiplier to 3x instead of 2x.  This one is kind of arbitrary but my thinking is that a bit more flexibility could improve load balancing and/or DRA scale up time.  It might not really matter.
- Fix tests that I broke in the process.

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -235,11 +235,8 @@ def _append_spark_config(spark_opts: Dict[str, str], config_name: str, config_va
     return spark_opts
 
 
-def _append_sql_shuffle_partitions_conf(spark_opts: Dict[str, str]) -> Dict[str, str]:
-    if 'spark.sql.shuffle.partitions' in spark_opts:
-        return spark_opts
-
-    num_partitions = 2 * (
+def _append_sql_partitions_conf(spark_opts: Dict[str, str]) -> Dict[str, str]:
+    num_partitions = 3 * (
         int(spark_opts.get('spark.cores.max', 0)) or
         int(spark_opts.get('spark.executor.instances', 0)) *
         int(spark_opts.get('spark.executor.cores', DEFAULT_EXECUTOR_CORES))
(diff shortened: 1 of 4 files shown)


service_configuration_lib/spark_config.py
```diff
@@ -252,22 +249,17 @@ def _append_sql_shuffle_partitions_conf(spark_opts: Dict[str, str]) -> Dict[str,
...
-        num_partitions_dra = 2 * (
+        num_partitions_dra = 3 * (
             int(spark_opts.get('spark.dynamicAllocation.maxExecutors', 0)) *
             int(spark_opts.get('spark.executor.cores', DEFAULT_EXECUTOR_CORES))
         )
         num_partitions = max(num_partitions, num_partitions_dra)
 
     num_partitions = num_partitions or DEFAULT_SQL_SHUFFLE_PARTITIONS
+    _append_spark_config(spark_opts, 'spark.sql.shuffle.partitions', str(num_partitions))
+    _append_spark_config(spark_opts, 'spark.sql.files.minPartitionNum', str(num_partitions))
```
88manpreet, line 260: I am not particularly sure if we should set this config by default.
I think this should be a lower value. Very low partition sizes is good for data-pipeline publishing but not good bunch of other actions, especially would probably take up very less executor resource, leading to higher cost.

I still incline towards `repartition` logic.

Side-note: We should still remove this, as it's defaults to `spark.default.parallelesim`
image

88manpreet (approved):
Left one optional comment. 
lgtm otherwise.

service_configuration_lib/spark_config.py
```diff
@@ -252,22 +249,17 @@ def _append_sql_shuffle_partitions_conf(spark_opts: Dict[str, str]) -> Dict[str,
         str(spark_opts['spark.dynamicAllocation.maxExecutors']) != 'infinity'
     ):
 
-        num_partitions_dra = 2 * (
+        num_partitions_dra = 3 * (
             int(spark_opts.get('spark.dynamicAllocation.maxExecutors', 0)) *
```
88manpreet, line 253: Maybe change this to `spark.dynamicAllocation.minExecutors`?

88manpreet: How about? 
if dra_enabled:
   num_partitions = min_executor_instances * cores * 2
else: 
   num_partitions = executor_instances * cores * 2

Otherwise, if you can add comment something in lines of, "We are inclining in the direction of more parallelism of tasks across max available cores"

88manpreet (approved):

PR #96, opened 2022-11-30
Fix pre-commit config

edingroot:
Fix flake8 source in pre-commit config which causes CI error.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,8 +20,8 @@ repos:
     rev: v1.4.3
     hooks:
     -   id: autopep8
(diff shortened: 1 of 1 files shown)


88manpreet (approved):
Nice. Not a blocker, just curious.
What was the error and do you know what wasn't it happening before?

edingroot:
> Nice. Not a blocker, just curious. What was the error and do you know what wasn't it happening before?

This error happens in the commits after this Nov:  

image  

I think it's because the flake8  repo on gitlab was either removed or made private, so I reference to pre-commit config in other yelp's repo for this change.

PR #99, opened 2022-12-05
Upgrade from py36 to py37 to fix CI test error

edingroot:
CI error:  
image

Reproduced and tested in devbox.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
-          python-version: 3.6
+          python-version: 3.7
       # GHA won't setup tox for us and we use tox-pip-extensions for venv-update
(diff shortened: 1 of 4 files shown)


88manpreet:
For this scope of change, please add someone from Compute Infra

edingroot:
Hi @nemacysts, can you please help to review the python version change of this repo? The jenkins job for updating internal pypi failed due to version conflict: https://jenkins-george.yelpcorp.com/blue/organizations/jenkins/mirrors-Yelp-service_configuration_lib-python/detail/mirrors-Yelp-service_configuration_lib-python/617/pipeline/64

nemacysts (approved):
this looks like it should be fine - seems like all the python packages that dep this one are already only testing on py37+, so this shouldn't negatively impact anyone if we start to use any py37+ features in this repo

nemacysts:
that said, we could also add some upper-bounds to the "problematic" dependencies if you don't want to raise the minimum required python version for this package

PR #100, opened 2022-12-13
MLCOMPUTE-548 | enabled DRA by default for Jupyter Spark sessions

CaptainSame:
- Moved default enable DRA logic from `paasta spark-run` to here.
- Set `minExecutors` equal to 0 by default for Jupyter Spark sessions.
- Set `initialExecutors` equal to min executors calculated using DEFAULT_DRA_MIN_EXECUTOR_RATIO (1/4) for Jupyter Spark sessions

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -301,7 +301,7 @@ def get_dra_configs(spark_opts: Dict[str, str]) -> Dict[str, str]:
         f'the cached data was lost, please consider increasing this value.\n',
     )
 
-    min_ratio_executors = 0
+    min_ratio_executors = None
     if 'spark.dynamicAllocation.minExecutors' not in spark_opts:
         # the ratio of total executors to be used as minExecutors
         min_executor_ratio = spark_opts.get('spark.yelp.dra.minExecutorRatio', DEFAULT_DRA_MIN_EXECUTOR_RATIO)
@@ -364,7 +364,7 @@ def get_dra_configs(spark_opts: Dict[str, str]) -> Dict[str, str]:
 
     # TODO: add regex to better match Jupyterhub Spark session app name
     if 'jupyterhub' in spark_app_name and 'spark.dynamicAllocation.initialExecutors' not in spark_opts:
-        if min_ratio_executors:
+        if min_ratio_executors is not None:
             # set initialExecutors default equal to minimum executors calculated above using
             # 'spark.yelp.dra.minExecutorRatio' and DEFAULT_DRA_MIN_EXECUTOR_RATIO for Jupyter Spark sessions
             initial_executors = min_ratio_executors
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -275,7 +275,7 @@ def get_dra_configs(spark_opts: Dict[str, str]) -> Dict[str, str]:
     ):
         return spark_opts
 
-    spark_app_name = spark_opts.get('spark.app.name', "")
+    spark_app_name = spark_opts.get('spark.app.name', '')
     # set defaults if not provided already
     _append_spark_config(spark_opts, 'spark.dynamicAllocation.shuffleTracking.enabled', 'true')
     _append_spark_config(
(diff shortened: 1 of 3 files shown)


service_configuration_lib/spark_config.py
```diff
@@ -307,20 +309,23 @@ def get_dra_configs(spark_opts: Dict[str, str]) -> Dict[str, str]:
                 min_executors, int(int(spark_opts['spark.dynamicAllocation.maxExecutors']) *
                                    float(min_executor_ratio)),
             )
+        # set minExecutors equal to 0 for Jupyter Spark sessions
+        if 'jupyterhub' in spark_app_name:
```
88manpreet, line 313: Add a log here in lines with: "Looks like you are launching spark session from jupyterhub notebook. Setting min executor count to 0, to save spark costs when any spark action is not running".

service_configuration_lib/spark_config.py
```diff
@@ -334,7 +339,20 @@ def get_dra_configs(spark_opts: Dict[str, str]) -> Dict[str, str]:
         log.warning(
             f'\nSetting spark.dynamicAllocation.maxExecutors as {max_executors}. If you wish to '
             f'change the value of maximum executors, please provide the exact value of '
-            f'spark.dynamicAllocation.maxExecutors in --spark-args\n',
+            f'spark.dynamicAllocation.maxExecutors in your spark args\n',
+        )
+
+    if 'jupyterhub' in spark_app_name and 'spark.dynamicAllocation.initialExecutors' not in spark_opts:
```
88manpreet, line 345: Is this the same logic in `paasta spark-run`?
Since we near end of year, would it better to make the change  in `spark-tools` to reduce the impact radius.
And then have a backlog ticket to have a common change in `service_configuration_lib` in early 2023?

service_configuration_lib/spark_config.py
```diff
@@ -334,7 +339,20 @@ def get_dra_configs(spark_opts: Dict[str, str]) -> Dict[str, str]:
...
+        # set initialExecutors default equal to spark.executor.instances for Jupyter
+        initial_executors = int(spark_opts.get('spark.executor.instances', DEFAULT_EXECUTOR_INSTANCES))
+        # initialExecutors shouldn't be more than maxExecutors
+        initial_executors = min(initial_executors, int(spark_opts['spark.dynamicAllocation.maxExecutors']))
```
88manpreet, line 349: Didn't we set initial executors as maxExecutors / 4?

service_configuration_lib/spark_config.py
CaptainSame: Adding

service_configuration_lib/spark_config.py
CaptainSame: I didn't understand. The condition check here that `spark_app_name` has `jupyterhub` is already reducing the impact radius

service_configuration_lib/spark_config.py
CaptainSame: Yes but there's a condition on line 345. This logic is only for Jupyter Spark sessions

service_configuration_lib/spark_config.py
88manpreet: In future version, we can have a slightly stricter regex check "^jupyterhub_.*".
Make sure to create a ticket to track, you can do it after pushing.

service_configuration_lib/spark_config.py
```diff
@@ -334,7 +359,24 @@ def get_dra_configs(spark_opts: Dict[str, str]) -> Dict[str, str]:
         log.warning(
             f'\nSetting spark.dynamicAllocation.maxExecutors as {max_executors}. If you wish to '
             f'change the value of maximum executors, please provide the exact value of '
-            f'spark.dynamicAllocation.maxExecutors in --spark-args\n',
+            f'spark.dynamicAllocation.maxExecutors in your spark args\n',
+        )
+
+    # TODO: add regex to better match Jupyterhub Spark session app name
+    if 'jupyterhub' in spark_app_name and 'spark.dynamicAllocation.initialExecutors' not in spark_opts:
+        if min_ratio_executors:
```
88manpreet, line 367: This is confusing.
`min_ratio_executors=min_executors` can also be `0`.
The better check and initialization probably is:
`min_executors = None`
if min_executors is not None:

CaptainSame: Okay. I can do that. Was just trying to keep the type same

88manpreet (approved):
lgtm.
I am assuming lots of manual tests have been done so that the status-quo things are not changing.
Please keep an eye and run tests after rollout.

PR #97, opened 2022-12-01
Cap executor resource with instance count adjustment

edingroot:
### Purpose of this PR
As indicated in MLCOMPUTE-533, this PR implemented the `Change Part-2`:  

Principles
1. Resources are determined based on memory requested by the user ("memory")
2. Keep memory:cpu = 7:1 to better fit aws instances
3. per_task_cpus = max(per_task_cpus, adjusted_executor_cpus)

Example:
- spark.executor.memory: 60GB, spark.executor.cores: 10, spark.executor.instances: 2
Total Current Memory: 60 * 2 = 120GB
- => Change to: spark.executor.memory: 56GB, spark.executor.cores: 8, spark.executor.instances: 2

### Testing
- Add unit tests
- Manually test run:  

1. Resources capped
```bash
paasta spark-run -s spark  --aws-credentials-yaml /etc/boto_cfg/mrjob.yaml --pool batch --spark-args "spark.executor.memory=200g spark.executor.cores=13 spark.executor.instances=2 spark.driver.memory=1g " --docker-memory-limit 1g --cmd "spark-submit /work/integration_tests/s3.py"
```
image

2. Resources adjusted
```bash
paasta spark-run -s spark  --aws-credentials-yaml /etc/boto_cfg/mrjob.yaml --pool batch --spark-args "spark.executor.memory=6g spark.executor.cores=2 spark.executor.instances=2 spark.driver.memory=1g " --docker-memory-limit 1g --cmd "spark-submit /work/integration_tests/s3.py"
```
image

DRA maxExecutors conf also overwritten:  
image

(with `--force-spark-resource-configs`)
image

3. Not adjusted - already 7:1
```bash
paasta spark-run -s spark  --aws-credentials-yaml /etc/boto_cfg/mrjob.yaml --pool batch --spark-args "spark.executor.memory=7g spark.executor.cores=1 spark.executor.instances=1 spark.driver.memory=1g " --docker-memory-limit 1g --cmd "spark-submit /work/integration_tests/s3.py"
```
image

4. Not adjusted - memory > multi-step release threshold 7g
```bash
paasta spark-run -s spark  --aws-credentials-yaml /etc/boto_cfg/mrjob.yaml --pool batch --spark-args "spark.executor.memory=28g spark.executor.cores=6 spark.executor.instances=1 spark.driver.memory=1g " --docker-memory-limit 1g --cmd "spark-submit /work/integration_tests/s3.py"
```
image

### Additional Notes
There needs to be an announcement that we are making this change.

PR #102, opened 2023-02-24
Release adjust spark executor resources for mem up to 14g

edingroot:
Follow up release of #97.  

Tested via `paasta spark-run` in devbox.

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -41,7 +41,7 @@ class ExecutorResourceConfig(NamedTuple):
 
 
 # Multi-step release - adjust if requested memory <= this threshold (memory in GB)
-ADJUST_EXECUTOR_MEM_CPU_RATIO_THRESH = 7
+ADJUST_EXECUTOR_MEM_CPU_RATIO_THRESH = 14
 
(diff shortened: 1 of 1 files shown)


service_configuration_lib/spark_config.py
```diff
@@ -41,7 +41,7 @@ class ExecutorResourceConfig(NamedTuple):
 
 
 # Multi-step release - adjust if requested memory <= this threshold (memory in GB)
-ADJUST_EXECUTOR_MEM_CPU_RATIO_THRESH = 7
+ADJUST_EXECUTOR_MEM_CPU_RATIO_THRESH = 14
```
nemacysts, line 44: i'd recommend moving this to srv-configs so that we don't need a version bump here and in paasta to update this value (and tbh, maybe the rest of the hardcoded values here to save us some work in the future)

nemacysts: we can always mock things out for tests - and in any case, we're already using srv-configs in this file: https://github.com/Yelp/service_configuration_lib/blob/4ee2fad539ff8c17a0e8f43f72a9dfdc72b34019/service_configuration_lib/spark_config.py#L28

edingroot: Thanks! Will do it with the next release.

PR #103, opened 2023-03-02
MLCOMPUTE-595 | remove DRA disable suggestion for notebooks

CaptainSame:
- Removed suggestion to disable DRA in notebooks
- Fixed cachedExecutorIdleTimeout warning (only show when not provided)
- Increase default cachedExecutorIdleTimeout to 25 mins. Add extra 15 mins in case of Jupyter
- Added more accurate cost calculations and warnings with DRA
- Move cost factors to srv configs
- Added nicer formatting
- Removed mesos tests

Testing:
image
image


edingroot (approved):

service_configuration_lib/spark_config.py
```diff
@@ -291,31 +291,38 @@ def get_dra_configs(spark_opts: Dict[str, str]) -> Dict[str, str]:
...
     log.warning(
-        'Spark Dynamic Resource Allocation (DRA) enabled for this batch. More info: y/spark-dra. '
-        'If your job is performing worse because of DRA, consider disabling DRA. To disable, '
-        'please provide spark.dynamicAllocation.enabled=false in your spark args\n',
+        'Spark Dynamic Resource Allocation (DRA) enabled for this batch. More info: y/spark-dra.\n',
     )
+    # TODO: add regex to better match Jupyterhub Spark session app name
+    if 'jupyterhub' not in spark_app_name:
+        log.warning(
+            'If your job is performing worse because of DRA, consider disabling DRA. To disable, '
```
88manpreet, line 302: Now that we have a successful mass rollout, easily available documentation, I don't think this message is needed anymore.

CaptainSame: Removed

service_configuration_lib/spark_config.py
```diff
@@ -291,31 +291,38 @@ def get_dra_configs(spark_opts: Dict[str, str]) -> Dict[str, str]:
...
 
-    spark_app_name = spark_opts.get('spark.app.name', '')
     # set defaults if not provided already
     _append_spark_config(spark_opts, 'spark.dynamicAllocation.enabled', 'true')
     _append_spark_config(spark_opts, 'spark.dynamicAllocation.shuffleTracking.enabled', 'true')
     _append_spark_config(
         spark_opts, 'spark.dynamicAllocation.executorAllocationRatio',
         str(DEFAULT_DRA_EXECUTOR_ALLOCATION_RATIO),
     )
+    if 'spark.dynamicAllocation.cachedExecutorIdleTimeout' not in spark_opts:
```
88manpreet, line 307: Cached executor idle timeout differs for different use-cases.
Jupyter notebook usage tends for longer time-gaps between two executive actions. 
It is better not clean cache for longer times
We have seen that increasing cached timeout for jupyter notebooks has helped in running their jobs faster without recomputation.

I suggest the following:
* Have a private api to identify whether a job is coming from Jupyterhub framework or not.
* Even better having an api which sets spark configs for Jupyterhub framework use-cases (e.g. min executors as 0, timeout as 45min+), log cost for jupyterhub notebooks
* Set 30 min cached timeout for regular tron jobs
* Set 45 min or higher cached timeout for jupyterhub jobs (with message of timeout value being set)

88manpreet:
We also need to add the log message for max possible cost per hour and per day for the resources users are requesting. 
If this part of this repo, let's add that as part of this PR.

service_configuration_lib/spark_config.py
```diff
@@ -980,6 +991,50 @@ def _convert_user_spark_opts_value_to_str(user_spark_opts: Mapping[str, Any]) ->
...
+    max_cores = per_executor_cores * (int(spark_conf.get("spark.executor.instances", DEFAULT_EXECUTOR_INSTANCES)))
+    min_cores = max_cores
+    if "spark.dynamicAllocation.enabled" in spark_conf and spark_conf["spark.dynamicAllocation.enabled"] == "true":
+        max_cores = per_executor_cores * (int(
+            spark_conf.get("spark.dynamicAllocation.maxExecutors", max_cores)
+        ))
+        min_cores = per_executor_cores * (int(
+            spark_conf.get("spark.dynamicAllocation.minExecutors", min_cores)
+        ))
+    if paasta_pool == "batch":
```
88manpreet, line 1005: We need to generalize this logic. 
Now we can have more pools beyond `batch` and `stable_batch`.

a) Create a global dict: Poo: Cost; Default -1 implying unknown or not calculated yet.
b) Read the pool list from puppet deployed config path

service_configuration_lib/spark_config.py
```diff
@@ -980,6 +991,50 @@ def _convert_user_spark_opts_value_to_str(user_spark_opts: Mapping[str, Any]) ->
...
+            TextColors.red(
+                TextColors.bold(
+                    f'\n!!! HIGH COST ALERT !!!',
+                )
+            )
+        )
+    if min_dollars != max_dollars:
+        log.warning(
+            TextColors.magenta(
+                f'\nThe requested resources are expected to cost between $ {TextColors.bold(str(min_dollars))} → '
```
88manpreet, line 1023: Change this to just highlight, "Up-to {max} cost per hour. And per day.".  No need for range.

PR #105, opened 2023-03-08
Add ability to do an assume-role-with-web-identity call for spark

Qmando:
This adds a new method to getting aws credentials, web identity. Web identity is used with kubernetes pod identity in particular, and this change will allow us to have paasta spark-run get a session and pass those credentials to the spark job.

This functionality depends on two environment variables being set, and if either is missing, it will emit a warning and then continue with the default boto3 session. It will also attempt to use the maximum 12 hours session but fall back to 1 hour with a warning, as the 12 hour maximum must be set in AWS for each role, 1 hour being the default max. Any other error will be raised. 

Added a unit test as well as performed some manual testing:
```
$ env | grep AWS
AWS_WEB_IDENTITY_TOKEN_FILE=/nail/home/qlo/token
AWS_ROLE_ARN=arn:aws:iam::528****:role/service_*****
$ python
>>> from service_configuration_lib import spark_config
>>> creds = spark_config.get_aws_credentials(assume_web_identity=True)
Session duration for arn:aws:iam::528****:role/service_*** is only 1 hour due to the role's maximum session duration
>>> creds
('ASIAXW****', 'TmZG****', 'FwoG*****')
```

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -17,7 +17,6 @@ from typing import Tuple
 from urllib.parse import urlparse
 
 import boto3
-import botocore.exceptions
 import ephemeral_port_reserve
 import requests
 import yaml
@@ -127,6 +126,7 @@ def get_aws_credentials(
     session: Optional[boto3.Session] = None,
     aws_credentials_json: Optional[str] = None,
     assume_web_identity: bool = False,
+    session_duration: int = 43200,
 ) -> Tuple[Optional[str], Optional[str], Optional[str]]:
     """load aws creds using different method/file"""
     if no_aws_credentials:
@@ -138,7 +138,7 @@ def get_aws_credentials(
             creds = json.load(f)
         return (creds.get('accessKeyId'), creds.get('secretAccessKey'), None)
     elif assume_web_identity:
-        credentials = assume_aws_role_web_identity()
+        credentials = assume_aws_role_web_identity(session_duration)
         if credentials:
             return (credentials['AccessKeyId'], credentials['SecretAccessKey'], credentials['SessionToken'])
         else:
@@ -165,7 +165,9 @@ def get_aws_credentials(
     )
 
 
-def assume_aws_role_web_identity():
+def assume_aws_role_web_identity(
+    session_duration,
+) -> Dict[str, str]:
     """
     Checks that a web identity token is available, and if it is,
     get an aws session and return a credentials dictionary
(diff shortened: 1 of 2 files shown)


service_configuration_lib/spark_config.py
```diff
@@ -135,6 +137,15 @@ def get_aws_credentials(
         with open(aws_credentials_json, 'r') as f:
             creds = json.load(f)
         return (creds.get('accessKeyId'), creds.get('secretAccessKey'), None)
+    elif assume_web_identity:
+        credentials = assume_aws_role_web_identity()
+        if credentials:
+            return (credentials['AccessKeyId'], credentials['SecretAccessKey'], credentials['SessionToken'])
+        else:
+            log.warning(
+                'Tried to assume role with web identity but either '
```
nemacysts, line 146: nit:
```suggestion
                'Tried to assume role with web identity but neither '
```

service_configuration_lib/spark_config.py
```diff
@@ -154,6 +165,36 @@ def get_aws_credentials(
     )
 
 
+def assume_aws_role_web_identity():
```
nemacysts, line 168: would be nice to add a type for the retval here :)

nemacysts (approved):
lgtm - probably want to wait for the coreml folks to take a look before merging tho

service_configuration_lib/spark_config.py
Qmando: *Ultra Pedantry Warning* If one was set but the other wasn't "neither" would be wrong.

nemacysts: oh, i also just re-read this and saw a `not` that I didn't see before - so this is actually fine as-is and you can ignore me :p

Qmando: "Do you want your warning to be grammatically perfect or intuitive?"
"Yes"

service_configuration_lib/spark_config.py
```diff
@@ -154,6 +165,36 @@ def get_aws_credentials(
...
+            DurationSeconds=43200,
+        )
+    except botocore.exceptions.ClientError:
+        resp = client.assume_role_with_web_identity(
+            RoleArn=role_arn,
+            RoleSessionName=f'SparkRun-{timestamp}',
+            WebIdentityToken=token,
+            DurationSeconds=3600,
+        )
+        log.warning(f"Session duration for {role_arn} is only 1 hour due to the role's maximum session duration")
```
danielpops, line 194: You could consider just making the durationSeconds a parameter for the function, and letting the failure bubble up if an invalid value is chosen. In theory, a role's max session could be as low as 15 minutes.

nurdann (approved):

danielpops (approved):

service_configuration_lib/spark_config.py
```diff
@@ -135,6 +137,15 @@ def get_aws_credentials(
         with open(aws_credentials_json, 'r') as f:
             creds = json.load(f)
         return (creds.get('accessKeyId'), creds.get('secretAccessKey'), None)
+    elif assume_web_identity:
```
88manpreet, line 140: AFAIK from the offline discussion with qlo, this needs to be used when credentials path is not given.

I think this can be confusing for Spark Users in debugging who are giving service credentials path and assume_aws_identity solution. 

I would suggest put the logic of at most one of the options is given and fail otherwise.

88manpreet:
I am still not convinced that this logic will work on spark jobs deployed via tron or even in adhoc runs.
This may work in JupyterHub sessions.

However, having discussed in the meeting, it's ok to test it with spark test job I shared.

PR #107, opened 2023-03-15
Update spark.cores.max after resource adjustment

edingroot:
Update `spark.cores.max` if it is explicitly set and the value is smaller than `adjusted_instance` * `adjusted_cores`.

Test on devbox:   
`spark.cores.max` adjusted from 32 to 72:  
```bash
paasta spark-run -s spark --aws-credentials-yaml /etc/boto_cfg/mrjob.yaml --pool batch --spark-args "spark.executor.cores=2 spark.executor.memory=32g spark.executor.instances=16 spark.cores.max=32" --cmd "spark-submit /work/integration_tests/s3.py"
```
image
image


service_configuration_lib/spark_config.py
```diff
@@ -647,6 +647,10 @@ def _calculate_resources(
         'spark.executor.instances': str(executor_instances),
         'spark.task.cpus': str(task_cpus),
     })
+    if 'spark.cores.max' in user_spark_opts:
+        core_max = int(user_spark_opts.get('spark.cores.max', 0))
+        required_cores = executor_instances * executor_cores
```
88manpreet, line 652: Looking at the code, it seems like this is the last location where executor_instances are calculated. It should be ok then.

service_configuration_lib/spark_config.py
```diff
@@ -647,6 +647,10 @@ def _calculate_resources(
...
+        user_spark_opts['spark.cores.max'] = str(max(core_max, required_cores))
```
88manpreet, line 653: Why do we need to take max(given_max_cores, calculated_max_cores)?
Should we just override given_max_cores to calculated_max_cores, and WARN user to overridden value and ask them to remove from the given configs?

service_configuration_lib/spark_config.py
```diff
@@ -647,6 +647,10 @@ def _calculate_resources(
         'spark.executor.instances': str(executor_instances),
         'spark.task.cpus': str(task_cpus),
     })
+    if 'spark.cores.max' in user_spark_opts:
```
88manpreet, line 650: I think if `spark.cores.max` is not in user_spark_opts, we should still set it, right?
That way users can remove this.

service_configuration_lib/spark_config.py
edingroot: Sure, let me change to overwriting the value. For warning we already have one here: https://github.com/Yelp/service_configuration_lib/pull/107/files/5ce41f5b3deb428b35e19594d2f80ba0d4e7e0df#diff-7e57e6df6168e4083ee9a3b454025d48cb1f68c100d45dd92cf9056010cd5220R480

service_configuration_lib/spark_config.py
edingroot: I think the reason why we cannot just remove the config is because the `spark.executor.instances` is calculated from `spark.cores.max` (before this function) if `spark.executor.instances` is not specified. We would tend to be not adding it if it's not required.

88manpreet: > I think the reason why we cannot just remove the config is because the spark.executor.instances is calculated from spark.cores.max ..
Agree.

Should we still set the config, `spark.cores.max` internally in our code, if not set by user?

Reason case:
User is giving both `spark.cores.max` and `spark.executor.instances`. It's common in on-point queries, to ask user to remove `spark.cores.max` config or replace with `spark.executor.instances`. This is done to avoid confusion of inconsistent values of (`spark.executor.instances` * `spark.instances.cores`) and `spark.cores.max` or even as a standard practice. 
Sometimes users can still set use this config in their code leading to failure of job or inconsistent with `spark.executor.instances * spark.executor.cores`

Having this set internally (if not present), would give us freedom to remove this config in spark configs:
* without breaking code
* or without risk of inconsistent expectations 

Thoughts?

88manpreet: Ok, in that case, no strong preferences. 
Feel free to ignore this comment if doesn't looks useful to address in this case.

service_configuration_lib/spark_config.py
88manpreet: Nit: https://github.com/Yelp/service_configuration_lib/pull/107/files/5ce41f5b3deb428b35e19594d2f80ba0d4e7e0df#diff-7e57e6df6168e4083ee9a3b454025d48cb1f68c100d45dd92cf9056010cd5220R479

Replace warning message to: 
log.warning(
            'spark.cores.max is DEPRECATED. Replace with '
            'spark.executor.instances={executor_instances} in --spark-args and in your service code as "spark.executor.instances * spark.executor.cores" if used.',
        )

88manpreet (approved):
lgtm once nit comments are addressed

service_configuration_lib/spark_config.py
edingroot: I updated the message and change the logic a little bit in the latest commit to always show it, regardless of whether `spark.executor.instances` is set or not.

88manpreet (approved):

PR #111, opened 2023-04-06
Change assume_aws_role to use user credentials rather than  web identity token

Qmando:
The code that I'm modifying was added in #105 but after that, I realized that spark-run doesn't happen in the context of a k8s pod. That means this code is currently unused and changing behavior will not break anything.

I decided to try to switch this to using a static credentials on the host rather than a web identity token file. The code is roughly the same, just with that key difference. When paired with a change to paasta, I was able to verify this:

```
$ paasta spark-run --service spark --cluster pnw-devc --pool batch --assume-aws-role arn:aws:iam::52xxxxx:role/spark_test_role -C 'aws sts get-caller-identity'
...
Selected cluster manager: kubernetes

Setting docker memory and cpu limits as 2g, 1 core(s) respectively.
{
    "UserId": "AROAXXXXX:SparkRun-1680651893",
    "Account": "52xxxxx",
    "Arn": "arn:aws:sts::52xxxxx:assumed-role/spark_test_role/SparkRun-1680651893"
}
```

See also https://github.com/Yelp/paasta/pull/3584

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -172,9 +172,9 @@ def get_aws_credentials(
 
 
 def assume_aws_role(
-    role_arn,
-    session_duration,
-    key_file,
+    role_arn: str,
+    session_duration: int,
+    key_file: str,
 ) -> Dict[str, str]:
     """
     Checks that a web identity token is available, and if it is,
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -131,8 +131,9 @@ def get_aws_credentials(
     profile_name: Optional[str] = None,
     session: Optional[boto3.Session] = None,
     aws_credentials_json: Optional[str] = None,
-    assume_web_identity: bool = False,
-    session_duration: int = 43200,
+    assume_aws_role_arn: Optional[str] = None,
+    session_duration: int = 3600,
+    assume_role_user_creds_file: str = '/nail/etc/spark_role_assumer/spark_role_assumer.yaml',
 ) -> Tuple[Optional[str], Optional[str], Optional[str]]:
     """load aws creds using different method/file"""
     if no_aws_credentials:
@@ -143,14 +144,13 @@ def get_aws_credentials(
         with open(aws_credentials_json, 'r') as f:
             creds = json.load(f)
         return creds.get('accessKeyId'), creds.get('secretAccessKey'), None
-    elif assume_web_identity:
-        credentials = assume_aws_role_web_identity(session_duration)
+    elif assume_aws_role_arn:
+        credentials = assume_aws_role(assume_aws_role_arn, session_duration, assume_role_user_creds_file)
         if credentials:
             return credentials['AccessKeyId'], credentials['SecretAccessKey'], credentials['SessionToken']
         else:
             log.warning(
-                'Tried to assume role with web identity but neither '
-                'AWS_WEB_IDENTITY_TOKEN_FILE or AWS_ROLE_ARN was not found',
+                'Tried to assume role with web identity but something went wrong ',
             )
     elif service != DEFAULT_SPARK_SERVICE:
         service_credentials_path = os.path.join(AWS_CREDENTIALS_DIR, f'{service}.yaml')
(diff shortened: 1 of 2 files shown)


88manpreet:
Code changes look good to me.
* I am more curious about the external user-experience and end-to-end testing. 
Same test command to test: `time  paasta spark-run -s spark --pool batch     <new-flag>    --cmd 'timeout 1h spark-submit  /work/integration_tests/s3.py'`

a) Tron jobs to test: [time  paasta spark-run -s spark --pool batch     --aws-credentials-yaml /etc/boto_cfg/mrjob.yaml     --cmd 'timeout 1h spark-submit  /work/integration_tests/s3.py'](https://sourcegraph.yelpcorp.com/sysgit/yelpsoa-configs/-/blob/spark/tron-pnw-devc.yaml?L15)


b) I believe IIRC, this is intended for tron service code deployments only? 
How do I as service user (say spark in above example) know what value to set for `--assume-aws-role`? 
Steps to create the `spark_test_role`? 

c) If b) is correct, would every spark user need to set value for `session_duration`? 

d) What is the long-term plan for jobs taking longer than 12h (I believe the max time is 12h)?

Qmando:
Hi @88manpreet 

A)
After updating permissions on spark-test-role, I was able to successfully run the integration test. 
Status Quo (mrjob): It took 3:22 to run (https://fluffy.yelpcorp.com/i/Zr78t6sDtjPZk4q6K9tfn2G9ThTGD5xG.html)
This code (spark-test-role): 2:36 to run (https://fluffy.yelpcorp.com/i/1kjn7lj4Xcp72gMSfj37D4wVL65dldXz.html)
Note on that last fluffy: I forgot to add DeleteObject permissions so it ran through the whole job then failed at the end, but it's definitely working. The time variation is also just noise, since assuming a role takes a fraction of a second.

B) 
>  this is intended for tron service code deployments only?
Yes, this is for Tron only. Though it *could* work anywhere, as long as it has the spark_role_assumer creds.

> How do I as service user (say spark in above example) know what value to set?
It's the ARN of the role to use. This is identical to what you'd set under `iam_role` for pod identity. Fluffy includes the example argument.

> Steps to create the spark_test_role?
It would be similar to any other aws role, eg https://docs.google.com/document/u/1/d/1lzu5N0nq1t_DPoZuJ6HeZMsNkYhMF6JECvlIxClFfsA/edit or other runbooks. It would need to include SparkAccessDev/SparkAccessProd permissions, as the role needs to write to the spark event logs bucket, and also have a TrustRelationship setup with the spark_role_assumer user. Infrasec team must review all IAM role creation, so along with runbooks, we'd ensure these permissions are added.

C) I haven't yet decided on the default role duration. That default would come from Paasta and be adjustable. Roles by default are limited to one hour, but you can set them to allow up to 12 hours.

D) This is unfortunately a hard limitation on this system. We'd tell people that if there's any possibility the job will run for more than 12 hours, they would need to continue using --aws-credentials-yaml. Eventually, I'd like to have spark use *actual* pod identity, which allows session refreshing. I chose to accept this limitation because my understanding was that the vast majority of jobs do not run that long, and implementing true pod identity is a much larger undertaking.

service_configuration_lib/spark_config.py
```diff
@@ -131,8 +131,9 @@ def get_aws_credentials(
     profile_name: Optional[str] = None,
     session: Optional[boto3.Session] = None,
     aws_credentials_json: Optional[str] = None,
-    assume_web_identity: bool = False,
-    session_duration: int = 43200,
+    assume_aws_role_arn: Optional[str] = None,
+    session_duration: int = 3600,
+    assume_role_user_creds_file: str = '/nail/etc/spark_role_assumer/spark_role_assumer.yaml',
```
nemacysts, line 136: hmm, it's a little unfortunate that we're leaking a Yelp-only path here - but apart from providing no default or doing a bit of srv-configs shenanigans, I guess there's not much we could do otherwise

service_configuration_lib/spark_config.py
```diff
@@ -171,24 +171,24 @@ def get_aws_credentials(
     )
 
 
-def assume_aws_role_web_identity(
+def assume_aws_role(
+    role_arn,
     session_duration,
+    key_file,
```
nemacysts, line 177: nit:
```suggestion
    role_arn: str,
    session_duration: int,
    key_file: str,
```

nemacysts (approved):

PR #113, opened 2023-05-03
Fix service_configuration_lib dependency conflict bug

88manpreet:
Issue: https://github.com/Yelp/service_configuration_lib/actions/runs/4873917098/jobs/8694615557

Alternate solution: Upgrade botocore to newer version. 

Tests: Unit tests

Adding @luisp for visibility. I couldn't find Jon Lee's github user-id.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 importlib-resources==5.4.0
 pyinotify==0.9.6
 pyyaml >= 3.0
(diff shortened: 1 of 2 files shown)


edingroot (approved):

edingroot (approved):

chriskuehl:
If you upgrade pip (probably by removing venv-update and pip-custom-platform, which are both deprecated: https://github.com/Yelp/service_configuration_lib/blob/2cf020d65a9e4b9f7ce9e3722ea643e4ef57205a/tox.ini#L3-L4) you'll get a dependency resolver which can resolve these issues for you without manual pinning like this.

PR #115, opened 2023-05-06
Add support for --use-eks in PaaSTA

nemacysts:
paasta spark-run (when run with `--use-eks`) will require a different set of configuration: we'll be using a kubeconfig with the standard EKS setup rather than giving spark a URL to our masters + certs to auth with.

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -918,6 +918,7 @@ def _get_k8s_spark_env(
     volumes: Optional[List[Mapping[str, str]]],
     paasta_pool: str,
     service_account_name: Optional[str] = None,
+    include_self_managed_configs: bool = True,
 ) -> Dict[str, str]:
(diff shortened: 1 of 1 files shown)


service_configuration_lib/spark_config.py
```diff
@@ -935,7 +937,6 @@ def _get_k8s_spark_env(
         'spark.kubernetes.pyspark.pythonVersion': '3',
         'spark.kubernetes.container.image': docker_img,
         'spark.kubernetes.namespace': 'paasta-spark',
-        'spark.kubernetes.container.image.pullPolicy': 'Always',
```
nemacysts, line 938: this is unnecessary and will waste time/data since we're telling k8s to never use the cached version of the image. since we always include the full tag, there's no reason to skip the cache since we never re-use tags:)

jfongatyelp (approved):

PR #119, opened 2023-07-21
Disable decommission for non-spot pools

88manpreet:
Decommission feature makes sense only for spot pools.
There have been issues seen in the past for on-demand pools where the spark executors were running for long-time even if there are no active tasks, leading to hanging jobs

This has potentially been linked to decommission feature.

There shouldn't be any downside of disabling this feature for non-spot pools.

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -1109,6 +1109,10 @@ class SparkConfBuilder:
         else:
             raise UnsupportedClusterManagerException(cluster_manager)
 
+        # Enabling auto-decommission feature to move data around makes sense for only spot nodes
+        if 'batch' != paasta_pool and 'spark.decommission.enabled' in spark_conf:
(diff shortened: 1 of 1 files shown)


service_configuration_lib/spark_config.py
```diff
@@ -1057,6 +1057,10 @@ def get_spark_conf(
         spark_conf, cluster_manager, paasta_pool, force_spark_resource_configs,
     )
 
+    # Enabling auto-decommission feature to move data around makes sense for only spot nodes
+    if 'batch' != paasta_pool and 'spark.decommission.enabled' in spark_conf:
```
nemacysts, line 1061: this seems a tad brittle/hard to change - should this come from srv-configs/puppet instead?

additionally, wouldn't this disable the decommission feature for the spot gpu pools?

88manpreet: we don't have spot gpu pools as yet and no plans as such in near future. 
I think having them here for now if fine.

PR #121, opened 2023-07-25
Fix spark master assignment

edingroot:
This fixes the problem of Spark master address will not be assigned when a service account name is set due to a change from PR #115.

Tested by building package tar and installing in Jupyter server.


88manpreet (approved):
good catch. 
lgtm for merging.
But would still have it preferred reviewed by luisp when he is back.

I think this has impact only for jupyter servers.

nemacysts (approved):
whoops, thanks for fixing this!

PR #122, opened 2023-08-10
Pick spark ui port with preference and add user pod label

edingroot:
Tested via uninstalling & installing the packaged tar.gz in jupyterhub.
- The `"spark.yelp.com/user": <user>` pod label is added on Spark executor pods
- The ui port of the first Spark session (in the first notebook tab) will be 39091
- From the second Spark session: a random ephemeral port will be picked

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -275,7 +275,7 @@ def _get_k8s_spark_env(
     _paasta_cluster = _get_k8s_resource_name_limit_size_with_hash(paasta_cluster)
     _paasta_service = _get_k8s_resource_name_limit_size_with_hash(paasta_service)
     _paasta_instance = _get_k8s_resource_name_limit_size_with_hash(paasta_instance)
-    user = os.environ.get('USER', '')
+    user = os.environ.get('USER', '_unspecified_')
 
     spark_env = {
         'spark.master': f'k8s://https://k8s.{paasta_cluster}.paasta:6443',
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -174,7 +174,7 @@ def assume_aws_role(
     return resp['Credentials']
 
 
-def _pick_random_port(preferred_port: int = 0):
+def _pick_random_port(preferred_port: int = 0) -> int:
     """Return a random port. """
     return ephemeral_port_reserve.reserve('0.0.0.0', preferred_port)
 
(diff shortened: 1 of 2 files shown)


service_configuration_lib/spark_config.py
```diff
@@ -63,6 +63,7 @@
     'spark.kubernetes.executor.label.paasta.yelp.com/service',
     'spark.kubernetes.executor.label.paasta.yelp.com/instance',
     'spark.kubernetes.executor.label.paasta.yelp.com/cluster',
+    'spark.kubernetes.executor.label.spark.yelp.com/user',
```
88manpreet, line 66: Utility of this?

service_configuration_lib/spark_config.py
```diff
@@ -74,6 +75,7 @@
 
 SUPPORTED_CLUSTER_MANAGERS = ['kubernetes', 'local']
 DEFAULT_SPARK_RUN_CONFIG = '/nail/srv/configs/spark.yaml'
+PREFERRED_SPARK_UI_PORT = 39091
```
88manpreet, line 78: Reason for picking-up the specific UI port, instead of random? 
Is it because it's hard to know the random port beforehand?

service_configuration_lib/spark_config.py
```diff
@@ -172,11 +174,8 @@ def assume_aws_role(
     return resp['Credentials']
 
 
-def _pick_random_port(app_name):
+def _pick_random_port(preferred_port: int = 0):
```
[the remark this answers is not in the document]
88manpreet: Can 2 spark applications have same UI port?

service_configuration_lib/spark_config.py
edingroot: Replied in the slack thread

service_configuration_lib/spark_config.py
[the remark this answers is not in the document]
edingroot: yeah I checked there is no other places using this. No, it will pick another ephemeral port if the `preferred_port` has been used

service_configuration_lib/spark_config.py
edingroot: This port is specified by the pod annotation when creating a jupyter server / spark driver pod, to let Prometheus to know what the scrape endpoint is.
The reason of fixing this port instead to dynamically assign it & add pod annotation on-the-fly is because we don't need to give user permission for modifying pod info which may lead to security concerns.

service_configuration_lib/spark_config.py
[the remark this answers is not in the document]
edingroot: `ephemeral_port_reserve.reserve()` called `bind(ip, 0)` when the port is used.

Ref: https://man7.org/linux/man-pages/man7/ip.7.html

image

nemacysts, line 177: we might as well fully annotate the types here :)
```suggestion
def _pick_random_port(preferred_port: int = 0) -> int:
```

nemacysts (approved):

88manpreet (approved):

PR #124, opened 2023-08-17
Setting spark.app.id to a more intuitive name

edingroot:
1. Setting spark app id to same as the spark app name
2. Replacing `-`, `.`, `,` characters to `_` for app id

Test: test starting a new spark session and call the Spark's API at `http://<driver_ip>:<ui_port>/api/v1/applications`:  
```
[ {
  "id" : "jupyterhub_chi_test_spark_39091_1692232285",
  "name" : "jupyterhub_chi_test-spark_39091_1692232285",
  "attempts" : [ {
    "startTime" : "2023-08-17T00:31:28.283GMT",
    "endTime" : "1969-12-31T23:59:59.999GMT",
    "lastUpdated" : "2023-08-17T00:31:28.283GMT",
    "duration" : 1038136,
    "sparkUser" : "chi",
    "completed" : false,
    "appSparkVersion" : "3.2.4",
    "startTimeEpoch" : 1692232288283,
    "endTimeEpoch" : -1,
    "lastUpdatedEpoch" : 1692232288283
  } ]
} ]
```

PR #125, opened 2023-08-17
Update spark app id naming convention and limit length

edingroot:
Follow up PR of #124
- Trim app id to match k8s pod label length limit
- For non-jupyterhub jobs: use `<paasta_service>_<paasta_instance>_<timestamp>` format

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -1098,7 +1098,7 @@ class SparkConfBuilder:
         if is_jupyter:
             raw_app_id = app_name
         else:
-            raw_app_id = f'{paasta_service}_{paasta_instance}_{int(time.time())}'
+            raw_app_id = f'{paasta_service}__{paasta_instance}__{int(time.time()) % 10000}'
         app_id = re.sub(r'[\.,-]', '_', _get_k8s_resource_name_limit_size_with_hash(raw_app_id))
 
         spark_conf.update({
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -1089,12 +1089,17 @@ class SparkConfBuilder:
             # We want to make the app name more unique so that we can search it
             # from history server.
             app_name = f'{app_base_name}_{ui_port}_{int(time.time())}'
+        is_jupyter = _is_jupyterhub_job(app_name)
 
         # Explicitly setting app id: replace special characters to '_' to make it consistent
         # in all places for metric systems:
         # - since in the Promehteus metrics endpoint those will be converted to '_'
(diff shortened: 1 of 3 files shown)


service_configuration_lib/spark_config.py
```diff
@@ -1089,12 +1089,17 @@ def get_spark_conf(
...
             # from history server.
             app_name = f'{app_base_name}_{ui_port}_{int(time.time())}'
+        is_jupyter = _is_jupyterhub_job(app_name)
 
         # Explicitly setting app id: replace special characters to '_' to make it consistent
         # in all places for metric systems:
         # - since in the Promehteus metrics endpoint those will be converted to '_'
         # - while the 'spark-app-selector' executor pod label will keep the original app id
-        app_id = re.sub(r'[\.,-]', '_', app_name)
+        if is_jupyter:
```
88manpreet, line 1098: Maybe add a TODO comment here to also change for non-jupyter pods with the logic we discussed.

service_configuration_lib/spark_config.py
```diff
@@ -1089,12 +1089,17 @@ def get_spark_conf(
...
+            raw_app_id = app_name
+        else:
+            raw_app_id = f'{paasta_service}_{paasta_instance}_{int(time.time())}'
```
88manpreet, line 1101:
```suggestion
            raw_app_id = f'{paasta_service}__{paasta_instance}__{int(time.time())}'
```

88manpreet (approved):
Would be better to release the paasta version change for this early morning on LDN time, to catch any failing batches sooner.

service_configuration_lib/spark_config.py
edingroot: I was thinking that we only have 53 character available, this will make more app ids being trimmed. What would you think? Both are okay to me

edingroot: I saw from Tron web, there are already quite a few servce+instance names > 53 chars, even if the length of timestamp is not counted in

88manpreet: I don't think 2 characters matter here much, when trimming happens. But having double underscores gives us the value of separating instance and service names. Can app.id contain '.'? If so, you can set it as <service>.<instance>
Is the timestamp here in millis or seconds? I would have have timestamp in seconds (even every 10second is fine to save a digit but that can be confusing) or even "timestamp % 100,000 (or 10,000)" as it's mostly for uniqueness. 

Thoughts?

edingroot: We are replacing special characters `.`, `,`, `-` to `_` to make app_id consistent with Prometheus metric names.
The timestamp is in seconds. Modding the timestamp sounds good to me

PR #127, opened 2023-09-06
MLCOMPUTE-746 | Remove the option for explicitly specifying app id

edingroot:
We added the option for explicitly specifying Spark app id in #126 as a safety measure for rolling out the new app id naming convention.

Now the new naming convention has been fully rolled out without any problem and no one is currently using the option. So I'm going to remove the option to cleanup the logic and avoid problems caused by user explicitly set the id.

Related PR: https://github.com/Yelp/paasta/pull/3695


88manpreet (approved):
Nice.

PR #128, opened 2023-09-15
MLCOMPUTE-949 | Pick spark ui port from a preferred port range

edingroot:
Pick the Spark UI (API) port from a preferred port range, for scraping metrics from multiple Spark sessions in a single Jupyter server.

Tested in a jupyter server and launched 11 Spark sessions, the port of the 11th session is a random ephemeral port that is currently available on the host/pod.

image

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -1068,6 +1068,9 @@ class SparkConfBuilder:
             spark_app_base_name
         )
 
+        # Pick a port from a pre-defined port range, which will then be used by our Jupyter
+        # server metric aggregator API. The aggregator API collects Prometheus metrics from multiple
+        # Spark sessions and exposes them through a single endpoint.
         ui_port = int(
             (spark_opts_from_env or {}).get('spark.ui.port') or
             utils.ephemeral_port_reserve_range(
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -77,8 +77,6 @@ K8S_BASE_VOLUMES: List[Dict[str, str]] = [
 
 SUPPORTED_CLUSTER_MANAGERS = ['kubernetes', 'local']
 DEFAULT_SPARK_RUN_CONFIG = '/nail/srv/configs/spark.yaml'
-PREFERRED_SPARK_UI_PORT_START = 39091
-PREFERRED_SPARK_UI_PORT_END = 39100
 
 log = logging.Logger(__name__)
 log.setLevel(logging.INFO)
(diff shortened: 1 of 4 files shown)


service_configuration_lib/spark_config.py
```diff
@@ -78,7 +77,8 @@
 
 SUPPORTED_CLUSTER_MANAGERS = ['kubernetes', 'local']
 DEFAULT_SPARK_RUN_CONFIG = '/nail/srv/configs/spark.yaml'
-PREFERRED_SPARK_UI_PORT = 39091
+PREFERRED_SPARK_UI_PORT_START = 39091
+PREFERRED_SPARK_UI_PORT_END = 39100
```
edingroot, line 81: I'm thinking to move these in to srv-confgs, so the jupyter server api handler can also read them

edingroot: srv-configs #27930

service_configuration_lib/utils.py
```diff
@@ -28,3 +35,52 @@ def load_spark_srv_conf(preset_values=None) -> Tuple[Mapping, Mapping, Mapping,
     except Exception as e:
         log.warning(f'Failed to load {DEFAULT_SPARK_RUN_CONFIG}: {e}')
         raise e
+
+
+def ephemeral_port_reserve_range(preffered_port_start: int, preferred_port_end: int, ip='127.0.0.1') -> int:
```
nemacysts, line 40: there's a small typo in `preffered_port_start`: `preferred_port_start` is probably what we wanted :)

nemacysts, line 40: it'd be nice to upstream this into the ephemeral_port_reserve repo, but we can wait to see how this works for y'all's usecase first if you'd like :)

nemacysts, line 40: it might be worth documenting why we need to do this here - maybe replace the current docstring with the reasoning for why this exists and link to the upstream docstring?

nemacysts (commented):
minor comment re: typo/possible upstreaming, but otherwise seems fine :)

service_configuration_lib/utils.py
edingroot: Fixed. Thanks for point out

edingroot: Sure! Updated docstring

edingroot: Sure! I added comments explaining the purpose of adding this function and linked to the upstream repository for reference. Let's leave it here for now and test it out first

88manpreet (approved):

PR #129, opened 2023-09-27
Show warning message when spark app id is explicitly specified

edingroot:
## Description
Add `spark.app.id` to `NON_CONFIGURABLE_SPARK_OPTS` for showing warning message when it is specified.

This change shouldn't affect any existing logic since the app id is set (overwritten) here at this line: https://github.com/Yelp/service_configuration_lib/blob/c9aaf8e267aba3dbc19982ed8dfc041d0bc1083a/service_configuration_lib/spark_config.py#L1108

## Test
image

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -39,6 +39,7 @@ CLUSTERMAN_YAML_FILE_PATH = '/nail/srv/configs/clusterman.yaml'
 
 NON_CONFIGURABLE_SPARK_OPTS = {
     'spark.master',
+    'spark.app.id',
     'spark.ui.port',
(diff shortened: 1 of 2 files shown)


service_configuration_lib/spark_config.py
```diff
@@ -1051,7 +1052,6 @@ def get_spark_conf(
             If not provided, it uses cert files at {K8S_AUTH_FOLDER} to authenticate.
         :param force_spark_resource_configs: skip the resource/instances recalculation.
             This is strongly not recommended.
-        :param explcitly setting spark.app.id
```
edingroot, line 1054: Removing this broken docstring

88manpreet (approved):

PR #136, opened 2023-12-27
Remove debian targets from Makefile

nemacysts:
We haven't used this as a debian package in eons - let's just get rid of that infra entirely

Additionally, I did some luisp strong opinion things:
* PHONY markers directly above the targets
* removed non-standard targets
* removed venv-update/p-c-p
* fixed requirements files (deleted requirements.txt + unpinned dev deps)


cuza (approved):

Makefile
```diff
@@ -11,49 +11,13 @@
...
 # See the License for the specific language governing permissions and
 # limitations under the License.
-UID:=`id -u`
-GID:=`id -g`
-ITERATION=yelp1
-
-.PHONY: test tests coverage clean
-
-
+.PHONY: test
```
bobokun, line 14: curious, what does .PHONY do exactly?

PR #137, opened 2024-01-30
Allow get_aws_credentials to assume_role_with_web_identity

Qmando:
This change will allow the caller of get_aws_credentials to use assume_role_with_web_identity with a simple argument. This is ultimately to allow `paasta spark-run` to be used from within a Pod Identity context. The previous workaround was to create a AWS config file with a profile that did the same thing.

Pairs with https://github.com/Yelp/paasta/pull/3787 

Manual test, after setting ENV variables with a real token:
https://fluffy.yelpcorp.com/i/9nt6FK4T2dlSScKdVF53371flGRPS05F.html

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -114,6 +114,7 @@ def get_aws_credentials(
     assume_aws_role_arn: Optional[str] = None,
     session_duration: int = 3600,
     assume_role_user_creds_file: str = '/nail/etc/spark_role_assumer/spark_role_assumer.yaml',
+    use_default_session=False,
 ) -> Tuple[Optional[str], Optional[str], Optional[str]]:
     """load aws creds using different method/file"""
     if no_aws_credentials:
@@ -132,6 +133,8 @@ def get_aws_credentials(
             log.warning(
                 'Tried to assume role with web identity but something went wrong ',
             )
+    elif use_default_session:
+        session = Session()
     elif service != DEFAULT_SPARK_SERVICE:
         service_credentials_path = os.path.join(AWS_CREDENTIALS_DIR, f'{service}.yaml')
         if os.path.exists(service_credentials_path):
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -114,7 +114,7 @@ def get_aws_credentials(
     assume_aws_role_arn: Optional[str] = None,
     session_duration: int = 3600,
     assume_role_user_creds_file: str = '/nail/etc/spark_role_assumer/spark_role_assumer.yaml',
-    use_default_session=False,
+    use_web_identity=False,
 ) -> Tuple[Optional[str], Optional[str], Optional[str]]:
     """load aws creds using different method/file"""
     if no_aws_credentials:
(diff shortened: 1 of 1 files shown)


danielpops (approved):

service_configuration_lib/spark_config.py
```diff
@@ -132,6 +133,27 @@ def get_aws_credentials(
             log.warning(
                 'Tried to assume role with web identity but something went wrong ',
             )
+    elif use_web_identity:
+        token_path = os.environ.get('AWS_WEB_IDENTITY_TOKEN_FILE')
+        role_arn = os.environ.get('AWS_ROLE_ARN')
+        if not token_path or not role_arn:
+            log.warning('No web identity token file found.')
+            return None, None, None
```
danielpops, line 141: This case seems like it should raise instead of returning the `(None, None, None)` tuple and logging a warning. If you want it to use web_identity and it doesn't find the token or the role ARN, it will likely make it more difficult to diagnose why permission errors are happening (e.g. whomever is calling this may end up just defaulting to whatever standard resolution like the instance profile

nurdann (approved):

service_configuration_lib/spark_config.py
```diff
@@ -132,6 +133,26 @@ def get_aws_credentials(
...
+            RoleArn=role_arn,
+            RoleSessionName=f'{service}-session-{timestamp}',
+            WebIdentityToken=token,
+            DurationSeconds=session_duration,
+        )
+        return (
+            session['Credentials']['AccessKeyId'],
+            session['Credentials']['SecretAccessKey'],
+            session['Credentials']['SessionToken'],
+        )
```
88manpreet, line 155: I don't know how to add comment outside the scope of changes, so adding it here.

Maybe also catch the exception of FileNotFound error if user gives assume-aws-role option to suggest them to aws-profile instead. Something we were discussing yesterday.

88manpreet: I meant for this case: https://github.com/Yelp/service_configuration_lib/blob/0219e2fe0b4ef23d63ca2f5eb3df2a120446c500/service_configuration_lib/spark_config.py#L189

88manpreet:
We are soon going to migration our spark jobs onto driver on k8s. Which means:
* Tron jobs: Spark driver would run on k8s pod (not batch box)
* Adhoc jobs: Spark driver for now keeps running on given host. But in future it would be launched on k8s pod

Would the functional behavior of using the above flags (assume-aws-role or aws-profile or use-web-identity) remain the same? 
cc'ing: @CaptainSame  who can help you provide context and testing for driver on k8s.

nemacysts (approved):

Qmando:
> We are soon going to migration our spark jobs onto driver on k8s. Which means:
> 
> * Tron jobs: Spark driver would run on k8s pod (not batch box)
> * Adhoc jobs: Spark driver for now keeps running on given host. But in future it would be launched on k8s pod
> 
> Would the functional behavior of using the above flags (assume-aws-role or aws-profile or use-web-identity) remain the same?

Seems like there's potentially two things that need to happen:
- `include paasta_tools::spark_role_assumer` in kube node puppet module
- Spark driver pod needs to mount `/nail/etc/spark_role_assumer/`

CaptainSame:
Is this meant to solve the `adhoc` `paasta spark-run` use-case where `AWS_WEB_IDENTITY_TOKEN_FILE` and `AWS_ROLE_ARN` env variables are present? Does the already implemented `--assume-aws-role` method have some limitations? Can you please explain the use-cases

CaptainSame:
In our latest release with Spark drivers running on k8s, we have removed the usage of `paasta spark-run` for Spark batches running on tron in favor of a cleaner `spark submit` (The role assumption shenanigans are being done using a KUBECONFIG file). So I am not sure where this use-case fits

nurdann:
> Is this meant to solve the `adhoc` `paasta spark-run` use-case where `AWS_WEB_IDENTITY_TOKEN_FILE` and `AWS_ROLE_ARN` env variables are present? Does the already implemented `--assume-aws-role` method have some limitations? Can you please explain the use-cases

This is specifically to solve the issue of `spark-run` not picking up pod identity when running Jenkins pipeline. See for more details https://jira.yelpcorp.com/browse/SEC-18599

CaptainSame (approved):
Sure. This looks good to me

service_configuration_lib/spark_config.py
```diff
@@ -165,10 +186,11 @@ def assume_aws_role(
             creds_dict = yaml.load(creds_file.read(), Loader=yaml.SafeLoader)
             access_key = creds_dict['AccessKeyId']
             secret_key = creds_dict['SecretAccessKey']
-    except PermissionError:
+    except (PermissionError, FileNotFoundError):
         log.warning(
-            'If using spark-run as a human, you must manually export '
-            'AWS session credentials first. See y/spark-run-aws-role',
+            f'Tried to use {key_file} but it is not available. --assume-aws-role '
```
88manpreet, line 191: Nit: Maybe also add the type of exception received?

PR #139, opened 2024-02-08
MLCOMPUTE-1160 | consolidate tron and paasta logic to service config lib

CaptainSame:
Testing:
Paasta:
Config diff between older version and this version: https://fluffy.yelpcorp.com/i/p4hQp4kN0lmsCJ7zLn3n0bRdm0W451Kz.html

Jupyter:
Config diff between older version and this version:
https://fluffy.yelpcorp.com/i/s27lb1H0sVbzb3q7cqjCrM3PbpkgPDKD.html


service_configuration_lib/spark_config.py
```diff
@@ -347,24 +347,6 @@ def _get_local_spark_env(
     }
 
 
-def _get_k8s_resource_name_limit_size_with_hash(name: str, limit: int = 63, suffix: int = 4) -> str:
```
88manpreet, line 350: Adding docs: Moved to utils.

service_configuration_lib/spark_config.py
```diff
@@ -929,31 +911,44 @@ def _append_event_log_conf(
 
         if len(self.spark_srv_conf.items()) == 0:
             log.warning('spark_srv_conf is empty, disable event log')
+            spark_opts.update({"spark.eventLog.enabled": "false"})
```
88manpreet, line 914: why?

edingroot: `spark.eventLog.enabled` is default to false, perhaps we don't need this line if srv_config is empty?

service_configuration_lib/utils.py
```diff
@@ -85,3 +113,46 @@ def ephemeral_port_reserve_range(preferred_port_start: int, preferred_port_end:
...
+
+        NOTE: This function is the same as paasta/paasta_tools/kubernetes_tools.py
+    """
+    if len(name) > limit:
+        digest = hashlib.md5(name.encode()).digest()
+        hashed = base64.b32encode(digest).decode().replace('=', '').lower()
+        return f'{name[:(limit-suffix-1)]}-{hashed[:suffix]}'
+    else:
+        return name
+
```
edingroot, line 147: Are the functions above moved to here because they will be used in multiple different repos? I think all the current 3 run types `paasta spark-run`, `tron_tools`, and `spark_tools` should call `get_spark_conf()` in service-configuration-lib to make things consistent except logic specific to each run types. IIUC, both `pod template` and `k8s resource name limitation` for executor pods applies to all run types.

service_configuration_lib/spark_config.py
```diff
@@ -929,31 +911,44 @@ def _append_event_log_conf(
...
-            return spark_opts
+            except Exception as e:
+                log.warning('Failed to identify account ID, error: {}'.format(str(e)))
+                spark_opts['spark.eventLog.enabled'] = 'false'
+                return spark_opts
+
+            for conf in self.spark_srv_conf.get('environments', {}).values():
+                if account_id == conf['account_id']:
+                    spark_opts['spark.eventLog.enabled'] = 'true'
+                    spark_opts['spark.eventLog.dir'] = conf['default_event_log_dir']
```
edingroot, line 937: Currently we are reading `default_event_log_dir` from srv_configs and set `spark.eventLog.enabled`, `spark.eventLog.dir` programmatically. Since we don't support custom eventLog dir, can we move these two Spark configs to `mandatory_defaults` in srv_configs to be loaded together with other static Spark configs?

service_configuration_lib/spark_config.py
CaptainSame: I have added this to be explicit and not leave the state to the default behaviour, since the log clearly says we are disabling it. I can remove this if you feel strongly against it

service_configuration_lib/utils.py
CaptainSame: This logic was in paasta spark-run and didn't exist in service configuration lib. I have moved it here because this is common and we are calling `get_spark_conf` everywhere

service_configuration_lib/spark_config.py
CaptainSame: But this config is not static because it needs to identify `account_id` or `environment` to choose the correct one dynamically. We have only put static ones in `mandatory_defaults`

service_configuration_lib/utils.py
edingroot: ++ I guess the idea is not to import and call functions in `service_configuration_lib.utils` by other repos, i.e. these will be only used by this repo itself

service_configuration_lib/spark_config.py
edingroot: I see, both looks okay to me

service_configuration_lib/utils.py
CaptainSame: That's correct

edingroot (approved):

service_configuration_lib/spark_config.py
88manpreet: Maybe we can remove this whole section? I can't think of a reason on why to disable to event log hen spark-srv-conf is empty. I recall generally spark event log used to be disabled to avoid risk of leaking of secrets. Since that is not happening anymore, AFAIK, maybe we can remove this section?

88manpreet:
Did you mean to attach fluffy for the Jupyter config diff as an example?

88manpreet (approved):

PR #141, opened 2024-02-21
Add unit tests for utils

edingroot:
Add unit tests in order to pass the internal pypi ingestion pipeline.


88manpreet:
How is the range decided? Is there a thread or ticket for failures in internal pypi ingestion pipeline?

88manpreet (approved):

edingroot:
> How is the range decided? Is there a thread or ticket for failures in internal pypi ingestion pipeline?

* The range is predefined in srv-configs, and is loaded [here](https://github.com/Yelp/service_configuration_lib/blob/56464b6/service_configuration_lib/spark_config.py#L1072-L1075).
* Failure details added in MLCOMPUTE-1175

PR #142, opened 2024-02-27
MLCOMPUTE-967 | Spark config calculation should result in failure if srv-configs can't be read

CaptainSame:
- Spark config generation is leading to failures like KeyError in runtime in cases where srv-configs are not present. We should actively fail fast in such cases rather than wait for KeyError failures in runtime
- Remove obsolete `no_aws_credentials` arg from `get_aws_credentials` function

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -5,7 +5,6 @@ import logging
 import math
 import os
 import re
-import sys
 import time
 from typing import Any
 from typing import Dict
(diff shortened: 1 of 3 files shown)


service_configuration_lib/spark_config.py
```diff
@@ -397,6 +395,8 @@ def __init__(self):
             ) = utils.load_spark_srv_conf()
         except Exception as e:
             log.error(f'Failed to load Spark srv configs: {e}')
+            # should fail because Spark config calculation depends on values in srv-configs
+            sys.exit(1)
```
nemacysts, line 399: might be a little cleaner to throw an exception - `sys.exit()` outside of something like the driver for a script can be a little unexpected

service_configuration_lib/spark_config.py
```diff
@@ -1067,13 +1067,16 @@ def get_spark_conf(
...
+        try:
+            ui_port = int(
+                (spark_opts_from_env or {}).get('spark.ui.port') or
+                utils.ephemeral_port_reserve_range(
+                    self.spark_constants.get('preferred_spark_ui_port_start'),
+                    self.spark_constants.get('preferred_spark_ui_port_end'),
+                ),
+            )
+        except Exception:
+            ui_port = utils.ephemeral_port_reserve_range()
```
88manpreet, line 1079: Add a warning log stating the error and back-up option.

edingroot (approved):

PR #144, opened 2024-06-03
Update spark-run to no longer override --aws-profile

choww:
Current behaviour from `paasta spark-run` is that `--aws-profile` will be overwritten if `--service` is provided (e.g. aws credentials will be obtained from `/etc/boto_cfg/<service>.yaml` instead of from the specified AWS profile)

This has caused some confusion from users, therefore this PR updates the behaviour such that:
- if `--aws-profile` is provided, use that value
- in absence of `--aws-profile`, then try to get credentials from boto_cfg
- in absence of any credential information provided, use the `default` AWS user profile

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -148,6 +148,8 @@ def get_aws_credentials(
             session['Credentials']['SecretAccessKey'],
             session['Credentials']['SessionToken'],
         )
+    elif profile_name:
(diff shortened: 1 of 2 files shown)


Qmando (approved):
LGTM

nemacysts:
@choww this code is owned by ML Compute, so I've added a couple folks from that team

choww:
Thanks [nemacysts](https://github.com/nemacysts)!

chi-yelp (approved):

PR #145, opened 2024-07-02
SEC-19136 | DAR-2360 | clearly prefer boto3 session over aws profile before checking service boto creds

CaptainSame:
Fixing as part of: https://github.com/Yelp/paasta/pull/3901


chi-yelp (approved):
lgtm! might also need a person from security to review

choww (approved):

PR #147, opened 2024-08-29
MLCOMPUTE-1496 | change logging levels to reduce logs

CaptainSame:
- Changed logging levels for various resource recalculation messages to INFO and change the default logging level of the module to WARN


88manpreet (approved):

nemacysts (approved):
nice - i think the remaining bits will be making it so that adding `-v` in paasta validate/setup_tron_namespace shows the info-level logs :)

(but that'll happen in the paasta repo)

PR #148, opened 2024-09-09
[spark] Stop checking if volume to mount in k8s exists locally

nemacysts:
It's entirely possible (and not an error) to create a podspec that contains paths to volume-mount that do not exist wherever service_configuration_lib is being invoked from.

As long as we ensure that we don't double-add things (which is only a problem due to the way that the env var keys are constructed here), skipping the existence check should be fine.

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -223,13 +223,12 @@ def _get_k8s_docker_volumes_conf(
 
     for volume in k8s_volumes:
         host_path, container_path, mode = volume['hostPath'], volume['containerPath'], volume['mode']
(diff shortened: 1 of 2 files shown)


jfongatyelp (approved):

nemacysts:
I think https://github.com/Yelp/service_configuration_lib/blob/85e8e3899358823f9e40fcf3028123aed61f5e00/service_configuration_lib/spark_config.py#L380 is the only place where we might run an executor with these files missing - but the container runtime should be fine with that: it'll simply create empty folders on-host rather than crash

(if we want to avoid that, I can update just that callsite to check for file existence)

88manpreet (approved):

PR #151, opened 2024-10-09
MLCOMPUTE-2001 | Skip UI port selection, pod tpl generation, and logs for driver on k8s tron

chi-yelp:
Did the following changes to reduce unnecessary noise from Spark jobs in the Tron deployment logs:
- Allow passing in a constructer flag `is_driver_on_k8s_tron` to `SparkConfBuilder`
- When `is_driver_on_k8s_tron` is true:
  - Skip Spark argument deprecation warnings
  - Skip trying to pick an available ephemeral port (always use the first port for the driver pod)
  - Skip generating pod template

(There will be a follow up PR in the paasta repo)

Test: build and install the package in the paasta repo

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -22,6 +22,8 @@ from boto3 import Session
 
 from service_configuration_lib import utils
 from service_configuration_lib.text_colors import TextColors
+from service_configuration_lib.utils import EPHEMERAL_PORT_END
+from service_configuration_lib.utils import EPHEMERAL_PORT_START
 
 AWS_CREDENTIALS_DIR = '/etc/boto_cfg/'
 AWS_ENV_CREDENTIALS_PROVIDER = 'com.amazonaws.auth.EnvironmentVariableCredentialsProvider'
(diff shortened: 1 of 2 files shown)


service_configuration_lib/spark_config.py
```diff
@@ -1105,20 +1112,22 @@ def get_spark_conf(
...
-                    self.spark_constants.get('preferred_spark_ui_port_end'),
-                ),
-            )
-        except Exception as e:
-            log.warning(
-                f'Could not get an available port using srv-config port range: {e}. '
-                'Using default port range to get an available port.',
-            )
-            ui_port = utils.ephemeral_port_reserve_range()
+        ui_port: int = self.spark_constants.get('preferred_spark_ui_port_start', EPHEMERAL_PORT_START)
```
nemacysts, line 1115: could probably use a comment here that for tron-launched drivers, the ui port will essentially be hardcoded (i'm assuming that we're ignoring `spark.ui.port` on purpose, so this comment would let future-us know that we didn't forget about that option)

nemacysts (approved):

service_configuration_lib/spark_config.py
chi-yelp: It seems we are still relying on `spark.ui.port` for tron-launched spark drivers (iirc the hard-coded one was removed a while ago), and the tron "executor" pod's ports are retrieved from [get_spark_ports_from_config](https://sourcegraph.com/github.com/Yelp/paasta@84cb7ecd911dc24fcadc5cd901c2430b163c4243/-/blob/paasta_tools/tron_tools.py?L1053-1059) if `executor: spark`.

chi-yelp: anyway I added some more comments for make it more clear

PR #154, opened 2025-02-27
Use C-accelerated safe_load for YAML

nemacysts:
We were generally already doing the right thing (calling load() with a CSafeLoader Loader), but this should make this a little more ergonomic (i.e., you just need to call safe_load() and not worry about passing in the right Loader to load())

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -7,6 +7,8 @@ omit =
     venv/*
     /usr/*
     setup.py
+    # this file has no real logic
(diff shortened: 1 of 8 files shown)


nemacysts:
(very similar to https://github.com/Yelp/paasta/pull/4011)

jfongatyelp (approved):

chi-yelp (approved):

PR #155, opened 2025-03-04
Add support for pod identity with executor

Qmando:
Paired with https://github.com/Yelp/paasta/pull/4016

A couple of changes to support launching spark jobs that have a service account/pod identity associated with the executors. 
- `spark.kubernetes.authenticate.executor.serviceAccountName` needs to be set in addition to the existing auth method. That one doesn't apply when the driver is local but executors are on k8s
- Change elif to if for the spark.master setting, as service_account_name now cannot be relied on to designate a tron run
- Switch to default provider instead of env provider, which is compatible with both env AND pod identity

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -27,6 +27,7 @@ from service_configuration_lib.utils import EPHEMERAL_PORT_START
 
 AWS_CREDENTIALS_DIR = '/etc/boto_cfg/'
 AWS_ENV_CREDENTIALS_PROVIDER = 'com.amazonaws.auth.EnvironmentVariableCredentialsProvider'
+AWS_DEFAULT_CREDENTIALS_PROVIDER = 'com.amazonaws.auth.DefaultAWSCredentialsProviderChain'
 GPU_POOLS_YAML_FILE_PATH = '/nail/srv/configs/gpu_pools.yaml'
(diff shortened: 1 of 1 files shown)


service_configuration_lib/spark_config.py
```diff
@@ -1125,8 +1127,8 @@ def get_spark_conf(
         spark_conf = {**(spark_opts_from_env or {}), **_filter_user_spark_opts(user_spark_opts)}
         random_postfix = utils.get_random_string(4)
 
-        if aws_creds is not None and aws_creds[2] is not None:
-            spark_conf['spark.hadoop.fs.s3a.aws.credentials.provider'] = AWS_ENV_CREDENTIALS_PROVIDER
+        if (aws_creds is not None and aws_creds[2] is not None) or service_account_name is not None:
+            spark_conf['spark.hadoop.fs.s3a.aws.credentials.provider'] = AWS_DEFAULT_CREDENTIALS_PROVIDER
```
Qmando, line 1131: `AWS_DEFAULT_CREDENTIALS_PROVIDER` will work with access keys in the env as well

nemacysts (approved):
probably wanna wait for someone on ml-compute to ship this since they own this, but lgtm

nurdann (approved):

chi-yelp (approved):

PR #156, opened 2025-03-21
MLCOMPUTE-4041 | Fix jupyter spark memory calculation with DRA

chi-yelp:
## Changes
- Support cost estimation for DRA-enabled Spark sessions
- Remove calculation logic of `spark.cores.max` - it's already parsed and converted to `spark.executor.instances`
- Added some refactor TODOs to be addressed together with Spark Configuration Service implementation

## Test
- Add/update unit tests
- Test in jupyterhub

Before:
```
---------------------------------------------------------------------------
Resource request (0 CPUs and 0GB memory total) estimated cost: 0.0 per day
---------------------------------------------------------------------------
```

After (numbers hidden as X.X):
```
------------------------------------------------------------------------------
Resource request (8 CPUs and 56.0GB memory total) estimated cost: X.X per day
------------------------------------------------------------------------------
```


service_configuration_lib/spark_config.py
```diff
@@ -1253,32 +1256,21 @@ def get_grafana_url(spark_conf: Mapping[str, str]) -> str:
 
 
 def get_resources_requested(spark_opts: Mapping[str, str]) -> Mapping[str, int]:
+    dra_enabled = spark_opts.get('spark.dynamicAllocation.enabled') == 'true'
```
timma-yelp, line 1259: is `true` always lower case or we need to check for `True` and others also?

service_configuration_lib/spark_config.py
```diff
@@ -1253,32 +1256,21 @@ def get_grafana_url(spark_conf: Mapping[str, str]) -> str:
...
-    memory_overhead: int = (
-        requested_memory
-        if requested_memory > 0
-        else max(384, int(0.1 * executor_memory))
-    )
-    total_memory = (executor_memory + memory_overhead) * num_executors
-    log.info(f'Requested total memory of {total_memory} MiB')
+    executor_memory_overhead = compute_requested_memory_overhead(spark_opts, executor_memory)
+    total_memory = (executor_memory + executor_memory_overhead) * num_executors
+    dra_enabled_string = ' (DRA enabled)' if dra_enabled else ''
```
timma-yelp, line 1271: `nit`: we don't need space in the prefix, we have space in the below `MiB {}`

tests/spark_config_test.py
```diff
@@ -1528,6 +1528,36 @@ def test_parse_memory_string(memory_string, expected_output):
     assert spark_config.parse_memory_string(memory_string) == expected_output
 
 
+@pytest.mark.parametrize(
+    'spark_opts,executor_memory,expected_output', [
+        # No memory overhead configured, default to 384 MB
+        ({}, 4096, 384),
+        # Memory overhead configured via spark.executor.memoryOverhead
```
timma-yelp, line 1535: `nit` Instead of comment, it's good to have a `id` for the test

timma-yelp (commented):
just minor comments, overall looks great to me

service_configuration_lib/spark_config.py
chi-yelp: It should be (lower case string), since these configs has been processed by get_spark_conf, but I also added `tolower` just to be safe

service_configuration_lib/spark_config.py
chi-yelp: updated, thanks!

tests/spark_config_test.py
chi-yelp: updated, but I personally find it hard to map between test cases lines with ids, hope adding comments same as ids helps

timma-yelp (approved):

service_configuration_lib/spark_config.py
```diff
@@ -1235,12 +1235,15 @@ def parse_memory_string(memory_string: Optional[str]) -> int:
         )
 
 
-def compute_requested_memory_overhead(spark_opts: Mapping[str, str], executor_memory):
-    return max(
+def compute_requested_memory_overhead(spark_opts: Mapping[str, str], executor_memory) -> float:
+    """Return memory overhead in MB."""
+    min_mem_overhead = 384
```
jhereth, line 1240: please add a comment where this magic number comes from (_why_ 384)

jhereth (changes requested):
looks good - just one comment (magic number)

service_configuration_lib/spark_config.py
chi-yelp: Added comments for the min 384mb, also for the default 10% overhead

jhereth (approved):

PR #163, opened 2025-06-03
Fix user check for jira ticketing

chi-yelp:
## Problem
paasta CLI commands such as `paasta validate`, `paasta mark-for-deployment` are blocked by the jira ticket check since they [call](https://github.com/Yelp/paasta/blob/master/paasta_tools/tron_tools.py#L353) the `get_spark_config` while trying the build the Tron configs.

More details in: https://github.yelpcorp.com/sysgit/srv-configs/pull/44869

## Solution
Use the `user` explicitly passed from get_spark_conf parameters for jira ticketing check, instead of reading from environment variables.

## Test
- Unit tests
- Hard-coding jira_ticket.enabled to true and build a whl and install to local paasta repo to test the following commands:
```
./.tox/py38-linux/bin/paasta validate -y ~/projects/sysgit/yelpsoa-configs -s spark
./tox/py38-linux/bin/paasta spark-run --aws-profile=dev --cmd bash
```

### DIFF
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -1060,7 +1060,7 @@ class SparkConfBuilder:
         user = user or os.environ.get('USER', None)
 
         if self.mandatory_default_spark_srv_conf.get('spark.yelp.jira_ticket.enabled') == 'true':
-            needs_jira_check = user not in TICKET_NOT_REQUIRED_USERS
+            needs_jira_check = cluster_manager != 'local' and user not in TICKET_NOT_REQUIRED_USERS
             if needs_jira_check:
                 valid_ticket = self._get_valid_jira_ticket(jira_ticket)
                 if valid_ticket is None:
--- a/service_configuration_lib/spark_config.py
+++ b/service_configuration_lib/spark_config.py
@@ -77,6 +77,9 @@ K8S_BASE_VOLUMES: List[Dict[str, str]] = [
 
 SUPPORTED_CLUSTER_MANAGERS = ['kubernetes', 'local']
 DEFAULT_SPARK_RUN_CONFIG = '/nail/srv/configs/spark.yaml'
+USER_BATCH = 'batch'  # used by batch servers
+USER_TRON = 'TRON'  # used by Tron jobs, or other paasta CLI commands such as `paasta validate/mark-for-deployment`
+USER_UNSPECIFIED = 'UNSPECIFIED'
 
 log = logging.Logger(__name__)
 log.setLevel(logging.WARN)
@@ -305,7 +308,7 @@ def _get_k8s_spark_env(
     service_account_name: Optional[str] = None,
     include_self_managed_configs: bool = True,
     k8s_server_address: Optional[str] = None,
-    user: Optional[str] = None,
+    user: Optional[str] = USER_UNSPECIFIED,
     jira_ticket: Optional[str] = None,
 ) -> Dict[str, str]:
     # RFC 1123: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names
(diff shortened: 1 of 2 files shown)


tests/spark_config_test.py
```diff
@@ -623,6 +623,20 @@ def gpu_pool(self, tmpdir, monkeypatch):
...
+                    'spark.executor.cores': '2',
+                    'spark.executor.instances': '600',
+                },
+                {
+                    'spark.executor.memory': '28g',
+                    'spark.executor.cores': '4',
+                    'spark.executor.instances': '600',
+                },
+                False,
+            ),
```
chi-yelp, line 639: for passing the test coverage > 90% check :p

chi-yelp, line 639: added this missing logic brach for passing the test coverage > 90% check :p

service_configuration_lib/spark_config.py
```diff
@@ -1051,8 +1053,11 @@ def get_spark_conf(
         # is str type.
         user_spark_opts = _convert_user_spark_opts_value_to_str(user_spark_opts)
 
+        # Get user from environment variables if it's not set
+        user = user or os.environ.get('USER', None)
+
         if self.mandatory_default_spark_srv_conf.get('spark.yelp.jira_ticket.enabled') == 'true':
-            needs_jira_check = os.environ.get('USER', '') not in ['batch', 'TRON', '']
+            needs_jira_check = user not in [USER_BATCH, USER_TRON, None]
```
nemacysts, line 1060:
```suggestion
            needs_jira_check = user not in TICKET_NOT_REQUIRED_USERS
```

service_configuration_lib/spark_config.py
```diff
@@ -77,6 +77,9 @@
 
 SUPPORTED_CLUSTER_MANAGERS = ['kubernetes', 'local']
 DEFAULT_SPARK_RUN_CONFIG = '/nail/srv/configs/spark.yaml'
+USER_BATCH = 'batch'  # used by batch servers
+USER_TRON = 'TRON'  # used by Tron jobs, or other paasta CLI commands such as `paasta validate/mark-for-deployment`
+USER_UNSPECIFIED = 'UNSPECIFIED'
```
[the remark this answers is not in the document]
nemacysts: also, I think this makes my next question a lot clearer: do we need to add UNSPECIFIED to this set?

service_configuration_lib/spark_config.py
```diff
@@ -1051,8 +1053,11 @@ def get_spark_conf(
...
 
+        # Get user from environment variables if it's not set
+        user = user or os.environ.get('USER', None)
+
         if self.mandatory_default_spark_srv_conf.get('spark.yelp.jira_ticket.enabled') == 'true':
-            needs_jira_check = os.environ.get('USER', '') not in ['batch', 'TRON', '']
+            needs_jira_check = user not in [USER_BATCH, USER_TRON, None]
             if needs_jira_check:
                 valid_ticket = self._get_valid_jira_ticket(jira_ticket)
                 if valid_ticket is None:
```
sidsatyelp, line 1063: Can we print out the user name in the `error_msg` as well. Will help debug outputs from different environments.

service_configuration_lib/spark_config.py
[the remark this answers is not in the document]
chi-yelp: It's for keeping the user pod label `spark.yelp.com/user` for the metrics / pod search, I changed the constant name to `USER_LABEL_UNSPECIFIED` for clarification

service_configuration_lib/spark_config.py
chi-yelp: Sure, added to the message

chi-yelp:
> How do we verify that this fixes the issue for `paasta mark-for-deployment` ? Afaik, that paasta subcommand has no dry-run option.

Perhaps @nemacysts will have a better idea. On the other hand, I think if `paasta validate` passes there shouldn't be a problem for `paasta mark-for-deployment`, as they construct tron configs in the same way

sidsatyelp (approved):

